PR #66, opened 2015-04-13
if the startup nodes cannot be connected, tell the users that the nodes ...

iandyh:
Currently if all the startup nodes are not reachable, it throws the exception with the message: All slots are not covered after query all startup_nodes..

However, it should tell users all the startup nodes are currently down instead.

Thanks!

### DIFF
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -80,10 +80,12 @@ class NodeManager(object):
         self.flush_slots_cache()
         all_slots_covered = False
         disagreements = []
+        startup_nodes_reachable = False
         for node in self.startup_nodes:
             try:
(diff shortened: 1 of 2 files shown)


iandyh:
The build failed with expectation. If you agree that this is the correct exception to be thrown, I will work on the test cases fixes too.

Thanks in advance.

Grokzen:
@iandyh The exception looks good to me. The build will still fail because of some flake8 things i have not got around to fix yet so you can ignore them if they are not on the lines that you have changed :]

iandyh:
Ok the test cases seem all passed except the ones with flake8.

iandyh:
ping @Grokzen

Grokzen:
@iandyh Thanks for this fix :]

PR #73, opened 2015-05-19
add redis cluster management commands: slots, info and nodes

iandyh:
A quick implementation.

The usage will be: `rcm = RedisClusterMgt(startup_nodes); rcm.info()`

There are some reasons behind the design: 

1. Since the class `RedisClusterMgt` will focus on cluster operation specific commands, it makes no sense to pass `cluster` command every time. Instead, directly calling `info` or `nodes` will be more concise. This does make newbies confusing because it is different from the official documentation. But they can grasp the concept quickly IMO.

2. Currently only three commands are implemented which are the ones related to meta data retrieval. The necessity of other types of commands is unknown since `redis-tri.rb` can do them well especially for complex commands like `addslots`

I am open for discussion. And if the current design is enough, I can work on some other commands and the tests.

Thanks!

### DIFF
--- a/rediscluster/cluster_mgt.py
+++ b/rediscluster/cluster_mgt.py
@@ -18,7 +18,7 @@ class RedisClusterMgt(object):
             startup_nodes=startup_nodes,
             init_slot_cache=True, **kwargs
         )
-        
+
     def __getattr__(self, attr):
         if attr in self.blocked_args:
             raise RedisClusterException('%s is currently not supported' % attr)
@@ -36,21 +36,21 @@ class RedisClusterMgt(object):
             finally:
                 self.connection_pool.release(c)
         return first_key(command, res)
-        
+
     def _execute_cluster_commands(self, *args, **kwargs):
         args = ('cluster',) + args
         node = self.connection_pool.nodes.random_node()
         return self._execute_command_on_nodes([node], *args, **kwargs)
 
     def info(self):
-        raw = self._execute_cluster_commands('info')        
+        raw = self._execute_cluster_commands('info')
         def _split(line):
             k, v = line.split(':')
             yield k
             yield v
         return {k: v for k, v in
                 [_split(line) for line in raw.split('\r\n') if line]}
-            
+
     def _make_host(self, host, port):
         return '%s:%s' % (host, port)
 
@@ -67,7 +67,7 @@ class RedisClusterMgt(object):
             for slave_ip, slave_port in slaves:
                 slave_host = nslookup(slave_ip) if host_required else slave_ip
                 slave_slots[self._make_host(slave_host, slave_port)].append(slots)
-        
+
         return {
             'master': master_slots,
             'slave': slave_slots
@@ -81,7 +81,7 @@ class RedisClusterMgt(object):
         return ret
 
     def nodes(self, host_required=False):
-        raw =  self._execute_cluster_commands('nodes')
+        raw = self._execute_cluster_commands('nodes')
         ret = {}
         for line in raw.split('\n'):
             if not line:
--- a/rediscluster/cluster_mgt.py
+++ b/rediscluster/cluster_mgt.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 from collections import defaultdict
 
-from .client import StrictRedisCluster
 from .connection import ClusterConnectionPool
 from .exceptions import RedisClusterException
 from .utils import clusterdown_wrapper, first_key, nslookup
--- a/rediscluster/cluster_mgt.py
+++ b/rediscluster/cluster_mgt.py
@@ -14,10 +14,10 @@ class RedisClusterMgt(object):
                     'getkeysinslot', 'keyslot', 'meet', 'replicate', 'reset',
                     'saveconfig', 'set_config_epoch', 'setslot', 'slaves')
 
-    def __init__(self, startup_nodes):
+    def __init__(self, startup_nodes=None, **kwargs):
         self.connection_pool = ClusterConnectionPool(
             startup_nodes=startup_nodes,
-            init_slot_cache=True            
+            init_slot_cache=True, **kwargs
         )
         
     def __getattr__(self, attr):
--- a/rediscluster/cluster_mgt.py
+++ b/rediscluster/cluster_mgt.py
@@ -44,6 +44,7 @@ class RedisClusterMgt(object):
 
     def info(self):
         raw = self._execute_cluster_commands('info')
+
         def _split(line):
             k, v = line.split(':')
             yield k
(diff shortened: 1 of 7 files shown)


rediscluster/cluster_mgt.py
```diff
@@ -0,0 +1,107 @@
...
+            'master': master_slots,
+            'slave': slave_slots
+        }
+
+    def _parse_node_line(self, line):
+        line_items = line.split(' ')
+        ret = line_items[:8]
+        slots = [sl.split('-') for sl in line_items[8:]]
+        ret.append(slots)
+        print ret
```
Grokzen: You need to write it with modern python print function statement, otherwise tests  will fail for newer versions of python.

```python
print(ret)
```

rediscluster/cluster_mgt.py
```diff
@@ -0,0 +1,107 @@
...
+        return ret
+
+    def nodes(self, host_required=False):
+        raw =  self._execute_cluster_commands('nodes')
+        ret = dict()
```
Grokzen:
```python
ret = {}
```

rediscluster/cluster_mgt.py
iandyh: Oops, I should remove that `print line`.

rediscluster/cluster_mgt.py
iandyh: The reason I don't like `{}` is that somehow it looks like a `set` to me but it is actually a `dict`. So `dict()` is preferred in my point of view although with minor performance sacrifice.

iandyh:
I'll remove the `print` tomorrow morning. 

I've looked at the implementation of redis-rb though, they implemented in the standard way which is `r.cluster('slots')`. Currently I don't know which is better. A specialised management class does make sense because in most of use cases they are not touched. 

Btw, what's the future plan for the project? Do you plan to merge it with redis-py in some days? Because it has already implemented most of the features and been quite stable.

Grokzen:
redis-py lib do not have any implemented cluster support right now. The only work i know that has been done is in this WIP PR 604 in redis-py. That work has been stalled out for now.

The future plans for this project is not set in stone right now because i was hoping that the work in that PR would continue and eventually be done and good enough so it could be merged into redis-py. 

Currently i do not have that much time to put into porting the entire lib myself back to redis-py.

The thing is that i do not know what else todo with this lib right now because it supports most of the stuff that a redis cluster can do in some way or another. What is needed is more people using it so bugs and problems can be identified and fixed. Other then that, all i can do is refactor the code more and more and bring it closer to the style redis-py already is using to make the porting easier.

One thing i know i want to do in the future is more exploration of pipelines and pubsub solutions because i know they need some work to see what can be done on the client side.

Grokzen:
@iandyh You have to fix the flake8 validation errors before i can merge it. But after that it is starting to look good :]

iandyh:
@Grokzen Good to hear! I will add tests to the module in the next few days.

Grokzen:
@iandyh Hi. Any progress on the tests for this code? Really want to merge it so i can make a new release containing this.

iandyh:
@Grokzen Hi, I just wrote tests for it yesterday. I'll submit them today for merge.

iandyh:
@Grokzen There are fail cases in the UT. However it seems like it was from previous code...

Grokzen:
The test failures is from your commits, you should be able to fix them by changing this code

```diff
diff --git a/tests/conftest.py b/tests/conftest.py
index 339fa14..a260f5e 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -76,17 +76,17 @@ def skip_if_redis_py_version_lt(min_version):
 @pytest.fixture()
 def o(request, **kwargs):
     """
-    Create a RedisCluster instance with decode_responses set to True.
+    Create a StrictRedisCluster instance with decode_responses set to True.
     """
-    return _init_client(request, decode_responses=True, **kwargs)
+    return _init_client(request, cls=StrictRedisCluster, decode_responses=True, **kwargs)
 
 
 @pytest.fixture()
 def r(request, **kwargs):
     """
-    Create a RedisCluster instance with default settings.
+    Create a StrictRedisCluster instance with default settings.
     """
-    return _init_client(request, **kwargs)
+    return _init_client(request, cls=StrictRedisCluster, **kwargs)
 
 
 @pytest.fixture()
```

tests/conftest.py
```diff
@@ -34,11 +34,11 @@ def _get_client(cls=None, **kwargs):
 
     params = {'startup_nodes': [{'host': '127.0.0.1', 'port': 7000}], 'socket_timeout': 10, 'decode_responses': False}
     params.update(kwargs)
-    return StrictRedisCluster(**params)
+    return cls(**params)
```
iandyh: @Grokzen Thanks for the fix. The UT fails because of this line and it was a bug before that providing a `cls` argument doesn't have any effect.

iandyh:
@Grokzen There seems to be an invocation error which relates to pyflakes?

Grokzen:
You have flake8 errors you have to fix, that is why you get a Invocation error

```
./rediscluster/__init__.py:8:1: F401 'RedisClusterMgt' imported but unused
from .cluster_mgt import RedisClusterMgt
^
./rediscluster/utils.py:99:1: W293 blank line contains whitespace
^
./rediscluster/utils.py:99:1: W391 blank line at end of file
^
./tests/test_redis_cluster_mgt.py:3:1: E302 expected 2 blank lines, found 1
class TestRedisClusterMgt(object):
^
./rediscluster/cluster_mgt.py:47:9: E301 expected 1 blank line, found 0
        def _split(line):
        ^
ERROR: InvocationError: '/home/travis/build/Grokzen/redis-py-cluster/.tox/flake8-py34/bin/flake8 --show-source --exclude=.venv,.tox,dist,docs,build,.git --ignore=E501,E731,E402 .'
```

You should be able to run flake8 locally with the following command if you have `tox` installed `tox -e flake8-py34` it will run the flake8 check on all code and you can see when you have fixed the issues. When you have done it i will be able to merge this.

Grokzen:
Oh yeah, one last thing. You must add a line in the file `CHANGES` describing what you have added. Also you can add yourself to `docs/Authors` if you want.

iandyh:
@Grokzen I am guessing this will be part of 0.3.0 release and I should add a line under it?

Grokzen:
Yes you can do that, You can also add `(iandyh)` at the end of the line to indicate that you made that contribution

Grokzen:
@iandyh Looks good now, i will have a look at it all later tonight and a release will be done in the next few days :]

iandyh:
@Grokzen Great! Thanks.

Grokzen:
Tested it out manually and works very well :]

A new release should be out very soon after i fix some other stuff.

Many thanks.

iandyh:
@Grokzen My pleasure. Please don't hesitate to @ me if bugs are found.

PR #76, opened 2015-06-28
Implemented READONLY mode

mumumu:
By default, Redis Cluster always returns MOVE redirection response on accessing slave node. To overcome this limitation for scaling read, I implemented [READONLY mode](http://redis.io/topics/cluster-spec#scaling-reads-using-slave-nodes) and its releated tests.

Please refer to docs/Readonly_mode.md for details.

### DIFF
--- a/tests/test_node_manager.py
+++ b/tests/test_node_manager.py
@@ -74,14 +74,26 @@ def test_init_slots_cache(s):
     Test that slots cache can in initialized and all slots are covered
     """
     good_slots_resp = [
-        [0, 5460, [b'127.0.0.1', 7000], [b'127.0.0.1', 7003]],
-        [5461, 10922, [b'127.0.0.1', 7001], [b'127.0.0.1', 7004]],
-        [10923, 16383, [b'127.0.0.1', 7002], [b'127.0.0.1', 7005]],
+        [0, 5460, [b'127.0.0.1', 7000], [b'127.0.0.2', 7003]],
+        [5461, 10922, [b'127.0.0.1', 7001], [b'127.0.0.2', 7004]],
+        [10923, 16383, [b'127.0.0.1', 7002], [b'127.0.0.2', 7005]],
     ]
 
     s.execute_command = lambda *args: good_slots_resp
     s.connection_pool.nodes.initialize()
     assert len(s.connection_pool.nodes.slots) == NodeManager.RedisClusterHashSlots
+    for slot_info in good_slots_resp:
+        all_hosts = ['127.0.0.1', '127.0.0.2']
+        all_ports = [7000, 7001, 7002, 7003, 7004, 7005]
+        slot_start = slot_info[0]
+        slot_end = slot_info[1]
+        for i in range(slot_start, slot_end + 1):
+            assert len(s.connection_pool.nodes.slots[i]) == len(slot_info[2:])
+            assert s.connection_pool.nodes.slots[i][0]['host'] in all_hosts
+            assert s.connection_pool.nodes.slots[i][1]['host'] in all_hosts
+            assert s.connection_pool.nodes.slots[i][0]['port'] in all_ports
+            assert s.connection_pool.nodes.slots[i][1]['port'] in all_ports
+
     assert len(s.connection_pool.nodes.nodes) == 6
 
 
@@ -300,12 +312,13 @@ def test_cluster_one_instance():
         }}
 
         assert len(n.slots) == 16384
-        assert n.slots[0] == {
-            "host": "127.0.0.1",
-            "name": "127.0.0.1:7006",
-            "port": 7006,
-            "server_type": "master",
-        }
+        for i in range(0, 16384):
+            assert n.slots[i] == [{
+                "host": "127.0.0.1",
+                "name": "127.0.0.1:7006",
+                "port": 7006,
+                "server_type": "master",
+            }]
 
 
 def test_init_with_down_node():
--- a/tests/test_node_manager.py
+++ b/tests/test_node_manager.py
@@ -79,20 +79,22 @@ def test_init_slots_cache(s):
         [10923, 16383, [b'127.0.0.1', 7002], [b'127.0.0.2', 7005]],
     ]
 
-    s.execute_command = lambda *args: good_slots_resp
-    s.connection_pool.nodes.initialize()
-    assert len(s.connection_pool.nodes.slots) == NodeManager.RedisClusterHashSlots
-    for slot_info in good_slots_resp:
-        all_hosts = ['127.0.0.1', '127.0.0.2']
-        all_ports = [7000, 7001, 7002, 7003, 7004, 7005]
-        slot_start = slot_info[0]
-        slot_end = slot_info[1]
-        for i in range(slot_start, slot_end + 1):
-            assert len(s.connection_pool.nodes.slots[i]) == len(slot_info[2:])
-            assert s.connection_pool.nodes.slots[i][0]['host'] in all_hosts
-            assert s.connection_pool.nodes.slots[i][1]['host'] in all_hosts
-            assert s.connection_pool.nodes.slots[i][0]['port'] in all_ports
-            assert s.connection_pool.nodes.slots[i][1]['port'] in all_ports
+    with patch.object(StrictRedis, 'execute_command') as execute_command_mock:
+        execute_command_mock.side_effect = lambda *args: good_slots_resp
+
+        s.connection_pool.nodes.initialize()
+        assert len(s.connection_pool.nodes.slots) == NodeManager.RedisClusterHashSlots
+        for slot_info in good_slots_resp:
+            all_hosts = [b'127.0.0.1', b'127.0.0.2']
+            all_ports = [7000, 7001, 7002, 7003, 7004, 7005]
+            slot_start = slot_info[0]
+            slot_end = slot_info[1]
+            for i in range(slot_start, slot_end + 1):
+                assert len(s.connection_pool.nodes.slots[i]) == len(slot_info[2:])
+                assert s.connection_pool.nodes.slots[i][0]['host'] in all_hosts
+                assert s.connection_pool.nodes.slots[i][1]['host'] in all_hosts
+                assert s.connection_pool.nodes.slots[i][0]['port'] in all_ports
+                assert s.connection_pool.nodes.slots[i][1]['port'] in all_ports
 
     assert len(s.connection_pool.nodes.nodes) == 6
 
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,6 @@
+* 1.1.0
+    * Added READONLY mode support, scales reads using slave nodes.
+
 * 1.0.0
     * No change to anything just a bump to 1.0.0 because the lib is now considered stable/production ready.
 
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ More detailed documentation can be found in `docs` folder.
 - [Pipelines](docs/Pipelines.md)
 - [Threaded Pipeline support](docs/Threads.md)
 - [Cluster Management class](docs/ClusterMgt.md)
+- [READONLY mode](docs/Readonly_mode.md)
 - [Authors](docs/Authors)
 
 
--- a/docs/Authors
+++ b/docs/Authors
@@ -18,3 +18,4 @@ Authors who contributed code or testing:
  - 72squared - https://github.com/72squared
  - Neuron Teckid - https://github.com/neuront
  - iandyh - https://github.com/iandyh
+ - mumumu - https://github.com/mumumu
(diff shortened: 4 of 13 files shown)


Grokzen:
Have you tried running a readonly command inside a pipeline and see what happens? It probably works but you should test it manually and see if it behaves as expected.

mumumu:
> Have you tried running a readonly command inside a pipeline and see what happens? It probably works but you should test it manually and see if it behaves as expected.

@Grokzen tested it manually, I think pipeline on READONLY mode works without using SET related operation, same as non-pipeline implementation. 

I added pipeline unit test code on READONLY mode. Please check it out.

Grokzen:
I will check it out and test it myself some during the weekend and prolly make a new minor release with this included :]

tests/test_node_manager.py
```diff
@@ -74,14 +74,26 @@ def test_init_slots_cache(s):
...
     s.execute_command = lambda *args: good_slots_resp
     s.connection_pool.nodes.initialize()
     assert len(s.connection_pool.nodes.slots) == NodeManager.RedisClusterHashSlots
+    for slot_info in good_slots_resp:
+        all_hosts = ['127.0.0.1', '127.0.0.2']
+        all_ports = [7000, 7001, 7002, 7003, 7004, 7005]
+        slot_start = slot_info[0]
+        slot_end = slot_info[1]
+        for i in range(slot_start, slot_end + 1):
+            assert len(s.connection_pool.nodes.slots[i]) == len(slot_info[2:])
```
Grokzen: This cause the following exception when running my redis cluster inside docker

```
s = RedisCluster<127.0.0.1:7000, 172.17.42.21:7000, 172.17.42.21:7001, 172.17.42.21:7002, 172.17.42.21:7003, 172.17.42.21:7004, 172.17.42.21:7005>

    def test_init_slots_cache(s):
        """
        Test that slots cache can in initialized and all slots are covered
        """
        good_slots_resp = [
            [0, 5460, [b'127.0.0.1', 7000], [b'127.0.0.2', 7003]],
            [5461, 10922, [b'127.0.0.1', 7001], [b'127.0.0.2', 7004]],
            [10923, 16383, [b'127.0.0.1', 7002], [b'127.0.0.2', 7005]],
        ]
    
        s.execute_command = lambda *args: good_slots_resp
        s.connection_pool.nodes.initialize()
        assert len(s.connection_pool.nodes.slots) == NodeManager.RedisClusterHashSlots
        for slot_info in good_slots_resp:
            all_hosts = ['127.0.0.1', '127.0.0.2']
            all_ports = [7000, 7001, 7002, 7003, 7004, 7005]
            slot_start = slot_info[0]
            slot_end = slot_info[1]
            for i in range(slot_start, slot_end + 1):
>               assert len(s.connection_pool.nodes.slots[i]) == len(slot_info[2:])
E               assert '172.17.42.21' in ['127.0.0.1', '127.0.0.2']
```

Grokzen: This diff should fix the problem i found here

```diff
diff --git a/tests/test_node_manager.py b/tests/test_node_manager.py
index 21add47..ed60f7a 100644
--- a/tests/test_node_manager.py
+++ b/tests/test_node_manager.py
@@ -79,20 +79,24 @@ def test_init_slots_cache(s):
         [10923, 16383, [b'127.0.0.1', 7002], [b'127.0.0.2', 7005]],
     ]
 
-    s.execute_command = lambda *args: good_slots_resp
-    s.connection_pool.nodes.initialize()
-    assert len(s.connection_pool.nodes.slots) == NodeManager.RedisClusterHashSlots
-    for slot_info in good_slots_resp:
-        all_hosts = ['127.0.0.1', '127.0.0.2']
-        all_ports = [7000, 7001, 7002, 7003, 7004, 7005]
-        slot_start = slot_info[0]
-        slot_end = slot_info[1]
-        for i in range(slot_start, slot_end + 1):
-            assert len(s.connection_pool.nodes.slots[i]) == len(slot_info[2:])
-            assert s.connection_pool.nodes.slots[i][0]['host'] in all_hosts
-            assert s.connection_pool.nodes.slots[i][1]['host'] in all_hosts
-            assert s.connection_pool.nodes.slots[i][0]['port'] in all_ports
-            assert s.connection_pool.nodes.slots[i][1]['port'] in all_ports
+    with patch.object(StrictRedis, 'execute_command') as execute_command_mock:
+        execute_command_mock.side_effect = lambda *args: good_slots_resp
+
+        s.connection_pool.nodes.initialize()
+
+        assert len(s.connection_pool.nodes.slots) == NodeManager.RedisClusterHashSlots
+
+        for slot_info in good_slots_resp:
+            all_hosts = ['127.0.0.1', '127.0.0.2']
+            all_ports = [7000, 7001, 7002, 7003, 7004, 7005]
+            slot_start = slot_info[0]
+            slot_end = slot_info[1]
+            for i in range(slot_start, slot_end + 1):
+                assert len(s.connection_pool.nodes.slots[i]) == len(slot_info[2:])
+                assert s.connection_pool.nodes.slots[i][0]['host'] in all_hosts
+                assert s.connection_pool.nodes.slots[i][1]['host'] in all_hosts
+                assert s.connection_pool.nodes.slots[i][0]['port'] in all_ports
+                assert s.connection_pool.nodes.slots[i][1]['port'] in all_ports
 
     assert len(s.connection_pool.nodes.nodes) == 6
```

The following command in `initialize` required mocking 

```
cluster_slots = r.execute_command("cluster", "slots")
```

Grokzen:
Another thing you need to do is to add a line into the changelog in some "next release" section with this new feature so that i can include that line in the next release when i set the next number. Also you can add yourself to the authors list if you like.

mumumu:
@Grokzen 

> This cause the following exception when running my redis cluster inside docker

Thank you for pointing out this test problem, I fixed it.

> Another thing you need to do is to add a line into the changelog in some "next release" section with this new feature so that i can include that line in the next release when i set the next number. Also you can add yourself to the authors list if you like.

Ok. updated CHANGES and docs/Authors.

Grokzen:
A question, i have been studying the PR that is in `redis-py` that tries to implement cluster support directly into the client, but i noticed one thing regarding `READONLY`. In that implementation https://github.com/andymccurdy/redis-py/pull/604/files#diff-7e2178bb22a1fed96812057da6cce88cR157 he is not using 2 different classes for readonly and not readonly, is there some reason why you have made 2 classes here and not merge them into 1?

I do not have a usecase or some realworld code to try these theories out and what is best way to use this feature but i think that maybe you want to be more precise when you want to use slave reads and maybe switch it back and forth without having to have 2 different clients, one for reading and one for writing.

@mumumu @72squared any thoughts and comments? I know you @72squared have more real world usage and knowledge then me and might have some thoughts.

72squared:
I don't have a strong real-world use case for read only ... My data set is not really read-heavy so I haven't needed to scale up reads by adding more slave instances. The masters have been able to handle the read/write load mix really well so far based on the size of my data set and my workload. 

As for how I'd probably prefer use an interface like this ... I usually have factory methods of instantiation so I'd just pass a flag into my factory method to get a read only instance so it makes more sense having different classes. If you use the same class for both and can change the behavior of the class to read-only in the middle of using the object then you might get some very weird behavior. Either an argument to the class constructor, or different classes is fine, as long as you can't change the behavior of the class after instantiation.

Grokzen:
Sounds like a immutable structure where you set it once and then it is fixed. I can live with that :p

Grokzen:
@mumumu I have been playing around with this some and the one thing that i noticed is that `READONLY` is sent to all nodes no matter if they are a master or slave. Should it not only be sent to slaves in the cluster and not master nodes?

Grokzen:
The only other thing is that if `READWRITE` command should be implemented in some function so that it can be used if someone needs it to restore the cluster according to the docs.

Grokzen:
Other than that it looks very good and mergable after the questions

Grokzen:
@mumumu Thanks for this. I might work and change some minor things within this feature in the future.

mumumu:
@Grokzen 

> A question, i have been studying the PR that is in redis-py that tries to implement cluster support directly into the client, but i noticed one thing regarding READONLY. In that implementation https://github.com/andymccurdy/redis-py/pull/604/files#diff-7e2178bb22a1fed96812057da6cce88cR157 he is not using 2 different classes for readonly and not readonly, is there some reason why you have made 2 classes here and not merge them into 1?

As for Connection class, I don't have a precise reason for having 2 classes of Connection and we can merge them into 1 class.

But  ConnectionPool must have 2 classes (ClusterConnectionPool, ClusterReadOnlyConnectionPool), because two different (readonly and non-readonly) Connection mode should be managed in different Connection Pool.  When we manage 2 kind of connection in one pool, implementation must be complicated.

> I do not have a usecase or some realworld code to try these theories out and what is best way to use this feature but i think that maybe you want to be more precise when you want to use slave reads and maybe switch it back and forth without having to have 2 different clients, one for reading and one for writing.

....

> The only other thing is that if READWRITE command should be implemented in some function so that it can be used if someone needs it to restore the cluster according to the docs.

My environment has almost read workload and I want to scale it as much as possible, so I implemented *immutable* readonly client and cannot switch it back with READWRITE command (This is a reason why I don't implement READWRITE command). If you swtich it back to normal connection client, we should instantiate StrictRedisCluster with readonly=False parameter.

mumumu:
> I have been playing around with this some and the one thing that i noticed is that READONLY is sent to all nodes no matter if they are a master or slave. Should it not only be sent to slaves in the cluster and not master nodes?

There is no problem when we send READONLY to all nodes because we can send set related operation to master with READONLY connection.

```
$ redis-cli -p 7000 cluster nodes | sort -k 2
473c2eaabd94a1bdaae6b82569c29d6e6d0c3868 127.0.0.1:7000 myself,master - 0 0 1 connected 0-5460
9af5e607cb44c79671e4292ad95ef37e0878b8f7 127.0.0.1:7001 master - 0 1437235731054 2 connected 5461-10922
7b7b43609ce3ac1be56f05c2c45f2d2412407fa4 127.0.0.1:7002 master - 0 1437235730051 3 connected 10923-16383
931cb6d7e36b916853651883db595b3b8b6cb2bf 127.0.0.1:7003 slave 473c2eaabd94a1bdaae6b82569c29d6e6d0c3868 0 1437235729050 4 connected
552286c8c325884c4ada547a2ecb158b9d1ffd3d 127.0.0.1:7004 slave 9af5e607cb44c79671e4292ad95ef37e0878b8f7 0 1437235731054 5 connected
da2eb13079edceebd0ca4a2aba82053fd31aa3ae 127.0.0.1:7005 slave 7b7b43609ce3ac1be56f05c2c45f2d2412407fa4 0 1437235729550 6 connected
$ echo -e "READONLY\r\nset b 1" | redis-cli -p 7000
OK
OK
$ echo -e "READONLY\r\nset b 1" | redis-cli -p 7003
OK
(error) MOVED 3300 127.0.0.1:7000
$ echo -e "READONLY\r\nget b" | redis-cli -p 7003
OK
"1"
```

mumumu:
>  Thanks for this. I might work and change some minor things within this feature in the future.

@Grokzen 

Thank you for merging this PR, but it seems that you have an interface problem about multiple functionality enabled connection class. IMO, we can have 1 connection class as said above. 

https://github.com/Grokzen/redis-py-cluster/pull/78#issuecomment-122059874

Grokzen:
Ye i think i will merge the 2 connection classes but keep the 2 pool classes as you stated.

Grokzen:
@mumumu I think i found a bug/issue with this and i am not sure right now how to fix it.

I am playing around with my own PR #77 and when i have moved some code around that handles a Moved response the new code breaks. I od not understand yet why the tests did not fail with your code but only started when i added my code.

The test that fails is `test_moved_redirection` with the following error

```python
self = <[KeyError("'host'") raised in repr()] SafeRepr object at 0x7f6ff0f23ea8>, slot = 12182

    def get_master_node_by_slot(self, slot):
>       return self.nodes.slots[slot][0]
E       KeyError: 0
```

The problem has it's roots in this function https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/client.py#L131 and methods called within it. The problem is that either the Moved handling was not working as expected before and i probably fixed it with my refactoring or else i do not really know O.o. It is almost like the moved code was not called or something. 

I think it is based here https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/client.py#L162 when the new node is set after a Moved redirection. One other thing is that should not all the slaves also be updated when a moved error is issues?

Grokzen:
Hmm. I might have spoken to soon about the error :p I found a minor typo that fixed the tests to pass for now, but i do think that the problem with not updating the slaves when Moved error is raised is still a issue and probably must be fixed. I will try and see if i can make up some test code that will fail and should pass after new code have been commited.

PR #78, opened 2015-07-16
Pub/private IP mapping support

dfb:
I tried to be surgical in the changes. The only change I dislike is the handling of the 'ssl' parameter, but the way I did it seems to most closely match the API of redis-py.

These changes support a couple of different use cases, but the main one that is driving it for me is the following:
- We have a redis cluster deployed in a data center somewhere, and the nodes talk to each other via private IP addresses on a private network.
- We need to be able to access this redis cluster from other data centers, but redis cluster doesn't really support any sort of authentication or security (redis itself has a tiny bit - the password that is passed in the clear - but it doesn't work with redis cluster, and hackery like command renaming doesn't currently work well with redis cluster - you can't rename the 'cluster' command for example).
- We run in lots of different data centers and are therefore reluctant to introduce a cross data center VPN if we can help it.
- SSL/TLS with a client certificate seems like the way to go - have a public IP for each redis node, but require the clients to use a secure connection and provide a certain client certificate - but redis doesn't support this yet.
- To get around this, we run a proxy on each redis node box that does SSL termination (e.g. stud/haproxy/whatever). Since redis-py already supports clients connecting via SSL, only minor changes were needed to redis-py-cluster to enable that as well.
- The missing piece is the IP address translation - redis knows only about the private IPs the nodes use to talk to each other, so when it sends back info on nodes/slots it tells the client about those IPs, which are different than the IPs the clients use. So this commit adds in optional support for remapping IPs from redis to the IPs the client uses.

Commit notes:
Added support for clients using different IPs to talk to redis cluster than what nodes use to talk to each other (e.g. public vs private IPs).

Added support for connecting via SSL (e.g. nodes are behind a proxy that does SSL termination).

Fixed potential problem where connection args weren't used by NodeManager.

### DIFF
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -16,10 +16,12 @@ from redis import ConnectionError
 class NodeManager(object):
     RedisClusterHashSlots = 16384
 
-    def __init__(self, startup_nodes=None):
+    def __init__(self, startup_nodes=None, host_map=None, **connection_kwargs):
         self.nodes = {}
         self.slots = {}
+        self.host_map = host_map or {}
         self.startup_nodes = [] if startup_nodes is None else startup_nodes
+        self.connection_kwargs = connection_kwargs
         self.orig_startup_nodes = [node for node in self.startup_nodes]
         self.pubsub_node = None
 
@@ -65,7 +67,7 @@ class NodeManager(object):
         return self.nodes[key]
 
     def get_redis_link(self, host, port, decode_responses=False):
-        return StrictRedis(host=host, port=port, decode_responses=decode_responses)
+        return StrictRedis(host=host, port=port, decode_responses=decode_responses, **self.connection_kwargs)
 
     def initialize(self):
         """
@@ -175,6 +177,7 @@ class NodeManager(object):
         """
         Update data for a node.
         """
+        host, port = self.host_map.get((host, port), (host, port)) # perform host/IP remapping if needed
         node_name = "{0}:{1}".format(host, port)
         self.nodes.setdefault(node_name, {})
         self.nodes[node_name]['host'] = host
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -7,7 +7,7 @@ import string
 import time
 
 # rediscluster imports
-from .connection import ClusterConnectionPool
+from .connection import ClusterConnectionPool, SSLClusterConnection
 from .exceptions import RedisClusterException, ClusterDownException
 from .pubsub import ClusterPubSub
 from .utils import (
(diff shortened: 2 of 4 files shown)


dfb:
Please ignore this PR until I figure out why the tests failed - sorry for the noise.

Grokzen:
One thing i noticed is that there is a new cluster connection class. In another PR where readonly is implemented it also will add a new connection class bringin it  up to 3 different classes. The problem is that you might want readonly and ssl enabled at the same time. With this way it is done right now it will be impossible to perform. I almost think that there should be only 1 class with settings like ssl and readonly that can be enabled when creating the cluster client instance.

dfb:
Agreed, but I wonder if this would be more appropriate as an upstream change? (to redis-py itself) Or maybe that's what you're referring to... anyway, redis-py has a normal connection and an SSL connection so I did the same here, but it'd be really great to not need all those classes floating around.

Grokzen:
Well after a closer look i do think that your SSL class should be there because it mirrors the structure already in redis-py. I guess you should continue to implement it this way until it is finished and working as expected. I will try to think some more about how to deal with readonly and this in a solution that offers both features.

Grokzen:
I think that one possible solution would be to take the SLLConnection class from redis-py and implement it here but then change the class it inherits from to not be Connection in redis-py but Cluster connection class in this lib. And that way readonly could be implemented in the cluster connection class and sslconnection class would inherit that property and we get both solutions. It is hackish but it could work in theory.

rediscluster/nodemanager.py
```diff
@@ -65,7 +67,7 @@ def random_node(self):
         return self.nodes[key]
 
     def get_redis_link(self, host, port, decode_responses=False):
-        return StrictRedis(host=host, port=port, decode_responses=decode_responses)
+        return StrictRedis(host=host, port=port, decode_responses=decode_responses, **self.connection_kwargs)
```
Grokzen: The problem with your tests is here. The connection args that is sent in `test_repr_contains_db_info_unix` test should not be sent down to StrictRedis i guess. The connection_args that goes into ClusterConnectionPool is not the same as the args that is going into ClusterConnection class, there must be some way to seperate them.

dfb: On Fri, Jul 17, 2015 at 3:07 PM, Grok <notifications@github.com> wrote:

> In rediscluster/nodemanager.py
> <https://github.com/Grokzen/redis-py-cluster/pull/78#discussion_r34931192>
> :
>
> > @@ -65,7 +67,7 @@ def random_node(self):
> >          return self.nodes[key]
> >
> >      def get_redis_link(self, host, port, decode_responses=False):
> > -        return StrictRedis(host=host, port=port, decode_responses=decode_responses)
> > +        return StrictRedis(host=host, port=port, decode_responses=decode_responses, **self.connection_kwargs)
>
> The problem with your tests is here. The connection args that is sent in
> test_repr_contains_db_info_unix test should not be sent down to
> StrictRedis i guess. The connection_args that goes into
> ClusterConnectionPool is not the same as the args that is going into
> ClusterConnection class, there must be some way to seperate them.
>
Yeah, I had fixed this in my branch (see
https://github.com/dfb/redis-py-cluster/commit/e3758258bd67ee47f8df0cea9bcfbb64d051d64e),
but then found other places where the tests are failing due to kwargs - for
example, one test passes 'foo' and 'bar' kwargs.

I'm going to pause working on my patch for awhile and come back to it if I
can think of a clean way to handle this.

Grokzen:
@dfb Btw, if you try again on this PR, could you possibly split it up into 2 different PR:s where one adds SSL support because i think that might be much easier to merge in because of the smaller number of changes, and then the host:port mapping feature in another PR?

dfb:
Yeah, I was hoping to go that route too, but I couldn't figure out a way to
do it really - since redis doesn't support SSL natively, the only way to
use SSL is to have some sort of proxy, and that's what then gets you to
needing the host:port mapping stuff. The only thing I could think of was
one PR that added the mapping stuff and then a later PR that added SSL -
that'd be doable.

PR #81, opened 2015-08-05
Add max_connections_per_node on ClusterConnectionPool

awestendorf:
This handles a situation where the max_connections argument applies to the whole cluster, and can starve connections on concurrent processes to nodes in the cluster. It also results in asymmetrical connection allocation across the nodes, which makes it challenging to configure file descriptor and redis max client settings.

The cluster-wide setting can also cause problems if the number of nodes is greater than max_connections.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@
     * Refactored exception handling and exception classes. 
     * Added READONLY mode support, scales reads using slave nodes.
     * Fix __repr__ for ClusterConnectionPool and ClusterReadOnlyConnectionPool
+    * Add max_connections_per_node parameter to ClusterConnectionPool so that max_connections parameter is calculated per-node rather than across the whole cluster.
 
 * 1.0.0
     * No change to anything just a bump to 1.0.0 because the lib is now considered stable/production ready.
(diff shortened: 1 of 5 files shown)


Grokzen:
@awestendorf This contains conflicts that needs to be resolved before i can merge this. Rebasing your commit ontop of my latest unstable should be enough.

Grokzen:
Other minor things is that you can add yourself to authors list and pleas add a line in the CHANGES document with a short description of the change.

awestendorf:
@Grokzen I saw that, will rebase and push the changes, add myself to the authors and changes too. Thanks.

Grokzen:
Btw was thinking about one thing, do you see any reason to keep the old way of tracking a global connection count or would this be better in all ways? I am thinking about deprecating the old way then and keep it until next release and remove it then.

One other minor thing. You should put a line in the docs/Upgrading.md document that it is recommended to change from the old implementation to this new one.

awestendorf:
I think the per-node feature is more useful, but I didn't want to change the API right away. A roll out like you described is probably best, but if you want to switch it over completely that works for us.

awestendorf:
@Grokzen I think all the changes are in place now

Grokzen:
I am going to play with it some before merging, should be done tomorrow i guess :]

Grokzen:
Thank you very much for this :]

PR #83, opened 2015-09-07
check pid before getting connection. Make fork safe

iandyh:
The current `get_connection_by_node` and `get_connection_by_slot` methods are not fork/thread safe.


Grokzen:
Looks good. Thanks for this :]

PR #86, opened 2015-09-29
Connection self-repair

iandyh:
It's a more sensible way to handle the connection error.

Before, the client will simply raise the exception when there is a network problem. Now it will retry once just like redis-py.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,9 +1,10 @@
 * Next release
-    * Refactored exception handling and exception classes. 
+    * Refactored exception handling and exception classes.
     * Added READONLY mode support, scales reads using slave nodes.
     * Fix __repr__ for ClusterConnectionPool and ClusterReadOnlyConnectionPool
     * Add max_connections_per_node parameter to ClusterConnectionPool so that max_connections parameter is calculated per-node rather than across the whole cluster.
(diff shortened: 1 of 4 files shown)


Grokzen:
The coverage tool have dropped support for python 3.2 in the latest release. https://coverage.readthedocs.org/en/latest/changes.html 

You must update the dev requirements file to install coverage >= 3.7.1, coverage < 4.0.0 to install a older supported version for now.It might be time to drop python 3.2 here also because i see that more and more :] But i will do that at a later time

Grokzen:
@iandyh Can you add a line in the CHANGES file with a short description of this fix? Something along the lines of "Improved error handling when sending commands? Is mergable after that is added.

iandyh:
@Grokzen Sure. I'll do it soon. This is a relatively quick fix to the problem. I actually want to reuse the connection reset part but it takes some time to think about it. I'll try to bring up in coming days.

Grokzen:
I will merge this for now? If you decide to make some other changes you can open a new PR by then.

iandyh:
@Grokzen Yes, please merge it.  :)

Grokzen:
Thanks for this fix :]

PR #91, opened 2015-10-08
Enable SCRIPT commands LOAD/FLUSH/EXISTS/EVALSHA

alisaifee:
Testing done
 * As per tests in [test_scripting.py](https://github.com/alisaifee/redis-py-cluster/blob/script-loading/tests/test_scripting.py)
 * Calling a registered script against a slot on a new node which doesn't have the script loaded.
 * Calling a registered csript on a slot that was moved to a different node which doesn't have the script loaded.

Not implemented
 * `SCRIPT KILL`
 * loading a script in a pipeline.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@
     * Add max_connections_per_node parameter to ClusterConnectionPool so that max_connections parameter is calculated per-node rather than across the whole cluster.
     * Improve thread safty of get_connection_by_slot and get_connection_by_node methods (iandyh)
     * Improved error handling when sending commands to all nodes, e.g. info. Now the connection takes retry_on_timeout as an option and retry once when there is a timeout. (iandyh)
+    * Added support for SCRIPT LOAD, SCRIPT FLUSH, SCRIPT EXISTS and EVALSHA commands. (alisaifee)
 
 * 1.0.0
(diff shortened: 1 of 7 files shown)


Grokzen:
@alisaifee You can add yourself to authors list/file. Also you need to add a line in the CHANGES file describing this change, add it in Next Release section. I guess also you need to start documenting the scripting solution with supported features, limitations and problems. One example of one thing that needs to be documented is the limitation that you must handle MOVED/ASK errors manually in the lua code and that it will not be provided within this library. I can continue to document things while testing scripting  if you start the work.

Grokzen:
Also are you planning to add more features or do you consider this code to be done?

alisaifee:
Regarding unimplemented features:
* I can't think of a safe implementation for `SCRIPT KILL` that is compliant with the implementation in `StrictRedis`. One possibility is to simply iterate over all masters and issue the command and return `True` if any node did have a script running, else raise the underlying exception from any of the nodes if none had any scripts running.
* Since there appears to be some work being done on pipelines in parallel, I feel it would be safer to hold off on that implementation for now.

I'll update authors/changes section and start work on the documentation for scripting support.

Grokzen:
LGTM :] Thank you very much for this

PR #94, opened 2015-10-23
handle ASK error responses correctly in pipeline context.

72squared:
better comments for logic in pipeline.


Grokzen:
Can you add a line in changes about this fix to ASK handling in pipelines & the improvment of doc strings?

72squared:
Added some better docs on pipelining, and the line in CHANGES

PR #100, opened 2015-10-28
removed assert from evaluation of dict, func: utils.is_dict.

gmolight:
Removing assert from utils.is_dict. Removing the risk of executing the passed down variable when assert  evaluates the variable. In addition set a default None return, since Nothing is already expected upstream.

### DIFF
--- a/rediscluster/utils.py
+++ b/rediscluster/utils.py
@@ -11,9 +11,7 @@ def is_dict(d):
     """
     Test if variable is a dict or not.
     """
-    if isinstance(d, dict):
-        return True
-    return None
+    return  isinstance(d, dict)
 
 
 def string_keys_to_dict(key_strings, callback):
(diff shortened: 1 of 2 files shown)


Grokzen:
Tests failed and would it not be better if you changed it to `return isinstance(d, dict)` ?

Grokzen:
Also, you would have to correct the unit tests so they would pass with this new logic becuas they are currently failing.

gmolight:
ok I will do so

gmolight:
sorry about that ..

Grokzen:
@gmolight I will go through this and review it sometime during the week

Grokzen:
@gmolight I implemented this fix in a different way then this but with the same result. The commit is 4738b6df9dd802a24ada8ff9920a3285306a33f3

I felt that raising a more proper exception on any error was better. Credit is till in your name even if i did the commit :]

Thanks for the idea and initial work.

PR #102, opened 2015-11-09
Password protected cluster fix

etng:
my redis cluster uses the following config to protected unauthenticated connection:
```
masterauth mypass
requirepass mypass
```
when connection it will throws the following exception:

```
ERROR sending 'cluster slots' command to redis server
```

this patch will fix the problem


Grokzen:
@etng Hi. You are on the right way to fix this. But you need to make a few changes before i can use this.

First you need to fix your branch so that it is based on the latest commit in unstable so i will not have 4 of my own commits merged again into unstable.

Secondly, the tests is failing for some reason and you need to investigate why that is and possibly fix the base cause of it or change the tests to fix them if they are broken.

Third, if there currently is no tests for passwords when connecting to the cluster, you should add some. If there is tests in `redis-py` package that does this already, you can copy them over from there to here and just include them.

Four, you can add yourself to authors list if you want. Please make a separate commit for this.

Five, please add a line in CHANGES file with a short description of what your fix is. Add in the existing or add a second called `Next release` in there. Please make a separate commit for this.

etng:
@Grokzen sorry for that, I am working on it.but there is a problem ,the  redis-trib.rb also not support password protected  connection, so I am working around to automation the setting up of password protected redis-cluster.can you give me some hint?

Grokzen:
@etng No i have no real answer to how to make it work in redis-trib.rb. I see you have 2 options to make it work maybe, either you send a fix to redis-trib.tb that includes the option to use username/password when dealing with the cluster. Or you have to build your cluster without username/password and after you have setup your cluster you then enable it manually and run with it. My suggestion is that you try to fix in redis-trib.rb because that will be alot easier i guess?

etng:
use https://github.com/Grokzen/redis-py-cluster/pull/102 instead

Issue #98, opened 2015-10-27
refactor pipeline send_cluster_commands


72squared:
This method is a huge unwieldy monster. If we get rid of threads it might shrink a bit but even still it'll be nasty. I'd like to take a stab at moving this logic into sub-tasks or possibly separate classes to ease understanding and breaking up the logic into more manageable chunks. However, I don't want to make any sacrifices on performance or break expectations on behavior so it'll need to be heavily tested.

Grokzen:
Closing because of merged PR

Issue #97, opened 2015-10-27
refactor pipeline threads to experiment with parallel write/read pattern


72squared:
We might not really need threads in pipelines if we first write to all the sockets for each node, sending the commands and wait to read the responses until all the writes have been sent first. This isn't perfect non-blocking i/o in the sense that you wouldn't be reading partial responses in parallel from the sockets, but in our case it may be good enough and in most cases might be even more performant than running a bunch of parallel threads, especially when the pipeline request maps to many many different nodes. Worth doing a prototype and building some performance benchmark comparisons.

Grokzen:
Closing because of merged PR

Issue #99, opened 2015-10-28
How to use pipeline when all commands are guaranteed to be in 1 slot/server ?


ddorian:
Hi,

I think sharding has to be considered really carefully, which I have, and my pipeline commands are guaranteed to go to one server/slot. Is there a way I can crate a pipeline where I guarantee that all commands will go to 1 server and no logic is executed that will group the commands by node? I saw docs and code but couldn't find anything.

Thank you

Grokzen:
Hi. There is no way to do this right now because the implementation is done to be generic and allow multi node/slot keys inside the same pipeline. But, to support this feature is something that i have been thinking about for a while, because it would rather easy to do by doing another class that inherits from the current pipeline class but replaces the execute function with a new function with your more basic logic. I have been thinking about doing something similar in the normal client for multi key commands. This is the way all other clients have implemented the client and pipelines. It would mean that it would provide a basic solution for both cases that have this way of dealing with keys. Then you would extend that functionality with a class that sits ontop of that basic class where all the new and fancy logic is implemented where it deals with multi node keys. With this design the user could choose what level of support and complexity it want.

Grokzen:
I will close this when the above referenced PR is closed. It improves the situation with pipelines alot and cuts alot of overhead when dealing with any number of nodes.

Grokzen:
Closing because of merged PR

PR #105, opened 2015-11-22
pipeline send commands to all nodes and then read sequentially

72squared:
addresses issues #97 and #98

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -96,7 +96,7 @@ class StrictRedisCluster(StrictRedis):
     )
 
     def __init__(self, host=None, port=None, startup_nodes=None, max_connections=32, init_slot_cache=True,
-                 pipeline_use_threads=True, readonly_mode=False, reinitialize_steps=None, **kwargs):
+                 readonly_mode=False, reinitialize_steps=None, **kwargs):
         """
         :startup_nodes:
             List of nodes that initial bootstrapping can be done from
@@ -106,8 +106,6 @@ class StrictRedisCluster(StrictRedis):
             Can be used to point to a startup node
         :max_connections:
             Maximum number of connections that should be kept open at one time
-        :pipeline_use_threads:
-            By default, use threads in pipeline if this flag is set to True
         :readonly_mode:
             enable READONLY mode. You can read possibly stale data from slave.
         :**kwargs:
@@ -141,7 +139,6 @@ class StrictRedisCluster(StrictRedis):
         self.nodes_callbacks = self.__class__.NODES_CALLBACKS.copy()
         self.result_callbacks = self.__class__.RESULT_CALLBACKS.copy()
         self.response_callbacks = self.__class__.RESPONSE_CALLBACKS.copy()
-        self.pipeline_use_threads = pipeline_use_threads
         self.reinitialize_counter = 0
         self.reinitialize_steps = reinitialize_steps or 25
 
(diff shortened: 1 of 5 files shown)


Grokzen:
@72squared I do think that this would also improve the situation for #99 because of the new way pipelines works. I would consider closing that issue also when this is merged.

rediscluster/client.py
```diff
@@ -96,7 +96,7 @@ class StrictRedisCluster(StrictRedis):
     )
 
     def __init__(self, host=None, port=None, startup_nodes=None, max_connections=32, init_slot_cache=True,
-                 pipeline_use_threads=True, readonly_mode=False, reinitialize_steps=None, **kwargs):
```
Grokzen: This change would require a note in the `docs/Upgrading.md` file stating that this argument can be removed from the implementers code.

rediscluster/client.py
```diff
@@ -154,7 +151,7 @@ def __repr__(self):
     def pubsub(self, **kwargs):
         return ClusterPubSub(self.connection_pool, **kwargs)
 
-    def pipeline(self, transaction=None, shard_hint=None, use_threads=None):
```
Grokzen: This change would require a note in the docs/Upgrading.md file stating that this argument can be removed from the implementers code.

rediscluster/client.py
```diff
@@ -850,7 +846,7 @@ class RedisCluster(StrictRedisCluster):
         }
     )
 
-    def pipeline(self, transaction=True, shard_hint=None, use_threads=None):
```
Grokzen: This change would require a note in the docs/Upgrading.md file stating that this argument can be removed from the implementers code.

Grokzen:
Also, could you add a line to the CHANGES file with this major change? Could you add some minor text like how much it could improve performance etc.

Grokzen:
@72squared I did some minor tweaks in `docs/pipeline.md` so you will have to rebase this PR on the latest code in unstable and fix the conflicts before i can merge this.

rediscluster/pipeline.py
```diff
@@ -534,22 +369,45 @@ def inner(*args, **kwargs):
...
+    def append(self, c):
+        self.commands.append(c)
 
-    def run(self):
+    def write(self):
+        """
+        Code borrowed from StrictRedis so it can be fixed
+        """
+        # build up all commands into a single request to increase network perf
+        connection = self.connection
```
Grokzen: Do we need local tmp variable here?

rediscluster/pipeline.py
```diff
@@ -534,22 +369,45 @@ def inner(*args, **kwargs):
...
-            self.value = self.execute(self.conn, self.cmds, False)
-        except Exception as e:
-            self.exception = e
+            connection.send_packed_command(connection.pack_commands([c.args for c in commands]))
+        except ConnectionError as e:
+            for c in commands:
+                c.result = e
+
+    def read(self):
+        connection = self.connection
```
Grokzen: Same here with local tmp variable

72squared: I did it as an optimization for the loop iteration. local variables are faster to access. But really it probably doesn't matter. Do you have a strong preference?

Grokzen: Hmm. Keep it as is.

Grokzen:
LGTM :]

PR #111, opened 2015-12-15
Allow passing in a custom connection pool

svrana:
Using this to support async cluster connections with greenlets and tornado. That code will live in http://github.com/svrana/gretis


Grokzen:
LGTM, thank you very much for this :]

72squared:
ooooh this looks sweet!

PR #114, opened 2015-12-29
make cluster down decorator work with instrumentation

72squared:
See: http://stackoverflow.com/questions/308999/what-does-functools-wraps-do

when instrumenting methods with this decorator, it reports the wrong method name. This patch should fix it. There might be other methods like this, but this is the most obvious one I saw.


Grokzen:
Very minor thing, could you change to `from functools import wraps` and `@wraps(func)`.

This way of doing it cause a unnecessary lookup at every use :]

Grokzen:
LGTM +1

PR #115, opened 2016-01-05
Remove pubsub implementation.

Grokzen:
- Drop all code that supports pubsub
 - Removed file rediscluster.pubsub
 - Updated docs to reflect the current pubsub support
 - client.pubsub, client.publish, client.subscribe, pipeline.publish, pipeline.subscribe
   now raises RedisClusterException.
 - Removed pubsub test code


ramonsnir:
Can you explain why you're doing this?

Grokzen:
This PR is no longer valid and have been superseded by the following commit f3b9c773838914ca4efe59e818b54b18cde6a511 that changed the behaviour of pubsub instead of removing it.

PR #117, opened 2016-01-06
Fix sdiffstore

mt3925:


Grokzen:
Hi @mt3925 
Could you please give a small example when/how it breaks so that i can understand why this is needed?

mt3925:
def sdiffstore(self, dest, keys, *args):
        """
        Store the difference of sets specified by ``keys`` into a new
        set named ``dest``.  Returns the number of keys in the new set.
        Overwrites dest key if it exists.
        Cluster impl:
            Use sdiff() --> Delete dest key --> store result in dest key
        """
        res = self.sdiff(keys, *args)
        self.delete(dest)

        if not res:
            return 0
        return self.sadd(dest, *res)
The difference of sets specified by ``keys`` may be empty, but the argument  ``res`` of the function ``sadd`` must be non-empty.

Grokzen:
Hi @mt3925. I tested this manually and created a failing test to verify this was broken and that this fixes and it does properly. I will add the test and a line in the CHANGES file after merging this.

PR #119, opened 2016-01-10
Fixed role in flags when there is more than 2 roles. Check for master or slave.

piront:


Grokzen:
LGTM, Thank you very much :]

PR #126, opened 2016-02-25
python 2.6 compatability

raven428:
python 2.6 compatability

### DIFF
--- a/rediscluster/cluster_mgt.py
+++ b/rediscluster/cluster_mgt.py
@@ -60,15 +60,12 @@ class RedisClusterMgt(object):
         return self._execute_command_on_nodes([node], *args, **kwargs)
 
     def info(self):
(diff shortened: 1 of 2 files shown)


Grokzen:
What is stated in this section https://github.com/Grokzen/redis-py-cluster#dependencies--supported-python-versions  but i have 0 plans on supporting python 2.6 what so ever in this code.

This might sound harsh, but if the code works for python 2.6 as is then fine, i will still not take any responsibility or try to maintain compatibility with 2.6 in all future code updates even through PR:s. There is a reason that i do not want to support python 2.6 and i do not really care about supporting it. If i could, i would drop python 2.7 but that is still far out. My suggestion for you is to move forwards to atleast the py2.7 but i know that it is very difficult for some companies to do this.

Issue #120, opened 2016-01-25
`StrictClusterPipeline` `raise_on_error` does not raise anything besides `ResponseError`.


leporid:
With regard to v1.1.0, I noticed some strange behavior when working on a client reconnection scheme.  Whereas I fully expected any `ConnectionError` to be raised from a pipeline execution, I instead quietly received an array of exactly as many `ConnectionError` instances as pipeline commands, with no exceptions raised.  There is a flag, `raise_on_error`, with default `True` in `StrictClusterPIpeline.execute()`, is passed along to `.send_cluster_commands()` (which also defaults the flag to `True`), and is finally used to determine whether or not to raise an exception before returning.

An exception is actually raised by `.raise_first_error()`, which has the form:

    def raise_first_error(self, commands, response):
        for i, r in enumerate(response):
            if isinstance(r, ResponseError):
                self.annotate_exception(r, i + 1, commands[i][0])
                raise r

It seems to me that an exception ultimately needs to be raised here if any of the response elements are instances of an appropriate child of `Exception`, whether the response element is a `ResponseError` or not.  For example, `redis.ConnectionError` is not raised.

Thanks!

72squared:
hmm, that's concerning. I haven't paid close attention to that section of the code in a while. I'll take a closer look.

72squared:
The issue is that raise_first_error is only looking for redis.exceptions.ResponseError but it should also raise on the ConnectionError.

https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/pipeline.py#L60-L65

Should be an easy fix. I'll submit a patch in a bit.

72squared:
probably should be changed to raise on any type of Exception since that type of object really isn't a valid return type object from redis cluster.

leporid:
That it should raise on any exception is what I assumed in making a temporary patch.  I can produce it for you, if that helps.

Thanks for looking into this.

72squared:
After thinking about it, we should look at the reference spec in redis-py to see how it implements the raise_on_error parameter:

https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L2492-L2551

Redis-py will still raise a connection error even if raise_on_error=False. It only suppresses them if the error is a ResponseError.

In our case, we are first trying to find other nodes in the cluster and giving the other nodes a chance to agree on a new master and run the command, but if all of that fails, we should still be raising a connection error no matter what.

[...]

PR #128, opened 2016-02-26
handle connection error issues in pipeline to address issue #120

72squared:


Grokzen:
@72squared I will trust you on this one :)

One thing tho, is this a fixer for the entire issue #120 or just a sub part of it?

72squared:
@Grokzen I did my best in this patch to follow the same behavior that the Redis Client does on pipelining. It raises an exception if it is a connection or timeout error regardless of whether or not the raise_error flag is passed, but all other exception types (ResponseErrors) respect the flag. This should bring us in line with that behavior.

Grokzen:
@72squared Sounds good. Many thanks for that patch :)

PR #130, opened 2016-03-14
Adding pub_sub_pool_size to control having connection leakages

### LINKED ISSUE #129: After Publish connections are not returned to the pool
When i use the pubsub mechanism inside redis, after a message gets published on a channel the connections tend to hog on one particular node in the cluster therby giving me this error
                       **Error 24 connecting to 127.0.0.1:7006. Too many open files.**
My cluster has 3 masters and 3 slaves.. Im attaching a small example script i used to test this behavior [example.py]. I checked the clients "redis-cli -c -p 7006 client list | wc -l"
Another peculiar thing i noticed was when i do 
for f in `seq 1 1000000`;do echo $f;redis-cli -c -p 7006 publish channel $f;done 
the connections are returned and 
[example.txt](https://github.com/Grokzen/redis-py-cluster/files/168839/example.txt)

no connection hogging is seen

melhin:
* Added a new optional parameter pub_sub_pool_size
* Check if length of _available_pubsub_connections < pub_sub_pool_size and then append
* On get_connection pop the connection from _available_pubsub_connections [if available else create a new connection]

This Closes #129

### DIFF
--- a/rediscluster/connection.py
+++ b/rediscluster/connection.py
@@ -66,7 +66,7 @@ class ClusterConnectionPool(ConnectionPool):
     RedisClusterDefaultTimeout = None
 
     def __init__(self, startup_nodes=None, init_slot_cache=True, connection_class=ClusterConnection,
-                 max_connections=None, max_connections_per_node=False, **connection_kwargs):
+                 max_connections=None, max_connections_per_node=False, pub_sub_pool_size = 500, **connection_kwargs):
         super(ClusterConnectionPool, self).__init__(connection_class=connection_class, max_connections=max_connections)
 
         self.max_connections = max_connections or 2 ** 31
@@ -78,6 +78,7 @@ class ClusterConnectionPool(ConnectionPool):
 
         self.connections = {}
         self.connection_kwargs = connection_kwargs
+        self.pub_sub_pool_size = pub_sub_pool_size
         self.reset()
 
         if "socket_timeout" not in self.connection_kwargs:
@@ -126,7 +127,10 @@ class ClusterConnectionPool(ConnectionPool):
             raise RedisClusterException("Only 'pubsub' commands can be used by get_connection()")
 
         # TOOD: Pop existing connection if it exists
-        connection = self.make_connection(self.nodes.pubsub_node)
+        try:
+            connection = self._available_pubsub_connections.pop()
+        except IndexError:
+            connection = self.make_connection(self.nodes.pubsub_node)
         self._in_use_pubsub_connections.add(connection)
 
         return connection
@@ -153,10 +157,11 @@ class ClusterConnectionPool(ConnectionPool):
         self._checkpid()
         if connection.pid != self.pid:
             return
-
         if connection in self._in_use_pubsub_connections:
             self._in_use_pubsub_connections.remove(connection)
-            self._available_pubsub_connections.append(connection)
+            ### Check max pub_sub_pool_size before appending so as to keep it within desired limits
+            if len(self._available_pubsub_connections) < self.pub_sub_pool_size:
+                self._available_pubsub_connections.append(connection)
         else:
             # Remove the current connection from _in_use_connection and add it back to the available pool
             # There is cases where the connection is to be removed but it will not exist and there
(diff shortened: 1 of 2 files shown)


melhin:
Test  408 failed because the initial cluster setup itself failed . Can you re run the build

Grokzen:
I will only merge this in order to rip it out in the other MR that will remove the pubsub support so that in the future when i need to add it back again, i know the source code and it includes this fix. Otherwise it is not worth to add in.

melhin:
Following up on this https://github.com/antirez/redis/issues/2672 
"""
In the current naive implementation every message is sent to every node, which is quite an overhead, but by limiting the number of instances to 3/4, the overhead is acceptable (the Pub/Sub cluster should only be used for Pub/Sub, basically, and if normal Redis operations are also required should be executed in a different cluster).
"""
By taking of pub sub completely we will not be able to use the cluster for pub sub. If people have 3/4 nodes they can still go ahead and use this naive implementation as the i/o cost is bearable in this case, so you can just leave a caveat saying this is known for increased overhead and the leave the implementation there. And when redis  fixes this issue antirez is saying by using bloom filters you can go ahead and change the implementation

ashishdisawal:
I was also stuck with this issue. You patch is working correctly.
Can't find any issue for now. Thanks for the time spent to fix it.

Grokzen:
@melhin you have conflicts that must be resolved before i can merge this PR. Please rebase it ontop of latest unstable and fix the conflicts.

Other minor things is that you should add a line in the CHANGES file about what you have added. Also please change the comments you added from `### ...` to the standards `# ...` comment syntax.

rediscluster/connection.py
```diff
@@ -66,7 +66,7 @@ class ClusterConnectionPool(ConnectionPool):
     RedisClusterDefaultTimeout = None
 
     def __init__(self, startup_nodes=None, init_slot_cache=True, connection_class=ClusterConnection,
-                 max_connections=None, max_connections_per_node=False, **connection_kwargs):
+                 max_connections=None, max_connections_per_node=False, pub_sub_pool_size = 500, **connection_kwargs):
```
Grokzen: Pep8 violation, should be `pub_sub_pool_size=500` see the other options on same row.

Grokzen:
@melhin One general question i have about this implementation is that if you put the limiting check on how many pubsub connections you can have when you put the connection back, i can come up with this scenario that would break that. Say that i set the limiting factor to 5 connections, i could then just hammer the create method to create new connections because there is no limit on how many i can create. This however would only be a problem if the connections would be long running and never released back to the pool. With this i could for example create 20 stuck connections that would exceed the limit of 5 and never have them release back to the pool. But it would not save more then the 5 first connections and after a while it would probably only reuse them.

I think you need to take a look at the existing method `make_connection` in the same class and implement a connection counter and block the connections at creation time.

melhin:
@Grokzen I understand what you are pointing out. But as far as what i know im actually looking at the connections that were created used and then released i.e available pool so my focus is on getting available pools limited. For long standing connections i haven't done any limiting . So i see there is an issue with more long standing connections how shall we go about it , make it flexible so that user decides on the active connection limit ? Shall we raise that as another issue and then look into it?

melhin:
@Grokzen I understand what you are pointing out. But as far as what i know im actually looking at the connections that were created used and then released i.e available pool so my focus is on getting available pools limited. For long standing connections i haven't done any limiting . So i see there is an issue with more long standing connections how shall we go about it , make it flexible so that user decides on the active connection limit ? Shall we raise that as another issue and then look into it?

Grokzen:
I can't accept this unless you change the behavior to what i proposed. Both `redis-py` https://github.com/andymccurdy/redis-py/blob/master/redis/connection.py#L936 and my connection pool class uses the same and correct pool limit schema https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/connection.py#L141 and the pubsub should follow the very same pattern.

Grokzen:
Note: PR #136 will superseed this PR

Grokzen:
PR #136 was merged into unstable branch that superseeds this so i will close this one

PR #132, opened 2016-03-20
Cluster could respond if redis doesn't require full coverage

baranbartu:
I have faced the issue while working on the company. One slot will be deleted somehow and cluster covered 16383 slots that it is not reachable according to redis-py-cluster. But redis has config to disable it using "cluster-require-full-coverage no" and we are using it. If even all slots are not covered then cluster still can be able respond. Currently each redis servers have the above config but we can not use it because of node manager all slots coverage check. With this PR, everyone will be able to use redis-py-cluster client with this config.

### DIFF
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -170,7 +170,7 @@ class NodeManager(object):
             # Validate if all slots are covered or if we should try next startup node
             for i in range(0, self.RedisClusterHashSlots):
                 if i not in tmp_slots:
-                    if not self.is_cluster_up():
+                    if self.cluster_require_full_coverage(nodes_cache):
                         all_slots_covered = False
 
             if all_slots_covered:
@@ -191,22 +191,23 @@ class NodeManager(object):
 
         self.determine_pubsub_node()
 
-    def is_cluster_up(self):
+    def cluster_require_full_coverage(self, nodes_cache):
         """
         if exists 'cluster-require-full-coverage no' config on redis servers,
         then even all slots are not covered, cluster still will be able to
         respond
         """
-        # get random node
-        node = self.orig_startup_nodes[0]
-        r = self.get_redis_link(host=node["host"], port=node["port"])
-        # get cluster info
-        cluster_info = r.execute_command("cluster", "info")
-        # parse cluster_state
-        for info in cluster_info.split("\r\n"):
-            if "cluster_state" in info:
-                return info.split(":")[1] == 'ok'
-        return False
+        nodes = self.nodes or nodes_cache
+
+        def node_require_full_coverage(node):
+            r_node = self.get_redis_link(host=node["host"], port=node["port"],
+                                         decode_responses=True)
+            return "yes" in r_node.config_get(
+                "cluster-require-full-coverage").values()
+
+        # at least one node should have cluster-require-full-coverage yes
+        return any(
+            node_require_full_coverage(node) for node in nodes.values())
 
     def determine_pubsub_node(self):
         """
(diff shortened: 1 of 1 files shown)


baranbartu:
I don't think my code is incompatible or something else that failed the builds. I have just seen same errors for all checks. What should i do? @Grokzen

Grokzen:
I can't see or figure out why this would break/stop. I tried to rerun the travis job but with the same output. I have not yet seen this in the unstable branch. I tried to rerun the `unstable` branch and i see no errors in there yet, for example this job: https://travis-ci.org/Grokzen/redis-py-cluster/jobs/112069136

I can dig in this later tonight to see what in your branch fails.

baranbartu:
Could you have a look what the problem is @Grokzen? Or i will delete and re-open pull request. 
This issue is really important to prevent SPOF even if one master and its all replica set are not reachable. IMHO.

Grokzen:
There is no need to make a new PR, i rerunned the travis-ci test manually and it had the same problem. Today is my last day before easter holiday and i will spend most of my free time working on redis-py-cluster because there is many standing issues that needs to be fixed and i really hope a new release can be made before next week because that has been in the making for a long time now :]

Grokzen:
@baranbartu I have been working some on this problem and i have pulled your branch locally and continued to work on it to fix the issues you found with your code. You can see the diff/commits here https://github.com/Grokzen/redis-py-cluster/compare/baranbartu-require-full-slot-coverage-no . I will notify in here when i merge it so you can see the end result of it.

baranbartu:
Thanks in advance for your interest. @Grokzen

Grokzen:
Btw, the problem i found was that you was calling `self.cluster_require_full_coverage(nodes_cache)` on every slot so that was why travis-ci got stuck because it it costly to create that many connections to the cluster and ask the same thing over and over and over again :)

baranbartu:
But i was calling it count of missing slots times? On my local env,there was one missing slot so it was called one time, but there are more than one missing slots on tests i guess. Anyway, you called it explicitly for one time like below and it is done right?  @Grokzen 

`need_full_slots_coverage = self.cluster_require_full_coverage(nodes_cache)`

Grokzen:
Ye, the tests had some really wierd scenario in it that caused some really wierd behaviour. If there was many missing slots from variable `tmp_slots` then it get stuck in this loop. Because the setting that you set in the redis server config is cluster wide and not specific to any slot, you only need to query it once and then you can reuse the value even if there is 0 slots in `tmp_slots` variable.

Grokzen:
Travis-ci now reports a OK test here https://travis-ci.org/Grokzen/redis-py-cluster/builds/119081336 so i will continue to merge this manually on my machine and push it to unstable branch.

Grokzen:
There we go, it was merged in this commit into unstable f1148ba
I also added a line in CHANGED and credited you with the fix :)
Clsoing this as fixed/implemented.

baranbartu:
Thanks:) So is there any plan for new release? @Grokzen

Grokzen:
There is still several pending issues and PR:s i want to process before i make a new release. But i want/need to get it out rather soonish because it has been a while sinc the last release now :)

baranbartu:
Well last question, i am not real contributor as far as i could understand. I won't be in the contributor list cause you have merged, right? little bit confused sorry:) @Grokzen

Grokzen:
Hmm, you will be in some way O.o Look here in the unstable branch and you can see that you are the author of 2 of the commits but it says that i was the commited of them because i tweaked them locally https://github.com/Grokzen/redis-py-cluster/commits/unstable

One thing you can do right away is to make a new PR and add your self to the file in `docs/Authors` and i merge it into unstable.

baranbartu:
I've opened new PR. Thanks again @Grokzen

PR #136, opened 2016-04-03
Rebuild how max connections is handled

Grokzen:
- max_connections_per_node is now a direct option in StrictRedisCluster and not part of kwargs option
 - pubsub connections is no longer tracked in seperate variables, now works same way ad redis-py handles pubsub connections
 - Counting and enforcing of max_connections and max_connections_per_node is no longer done with _in_use_connections
   but with the tracker of how many that has been created so far
 - Added threading test method 'test_pubsub_thread_publish' to more easily verify that hammering publish commands works as expected
   even in a threaded environment.

### DIFF
--- a/rediscluster/connection.py
+++ b/rediscluster/connection.py
@@ -106,10 +106,9 @@ class ClusterConnectionPool(ConnectionPool):
         """
         self.pid = os.getpid()
         self._created_connections = 0
+        self._created_connections_per_node = {}  # Dict(Node, Int)
         self._available_connections = {}  # Dict(Node, List)
         self._in_use_connections = {}  # Dict(Node, Set)
-        self._available_pubsub_connections = []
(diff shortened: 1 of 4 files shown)


rediscluster/connection.py
```diff
@@ -139,21 +138,31 @@ def get_connection(self, command_name, *keys, **options):
...
+        node = self.get_master_node_by_slot(slot)
 
-        # TOOD: Pop existing connection if it exists
-        connection = self.make_connection(self.get_master_node_by_slot(slot))
-        self._in_use_pubsub_connections.add(connection)
+        self._checkpid()
+
+        try:
+            connection = self._available_connections.get(node["name"], []).pop()
+            print("Getting existing connection")
```
Grokzen: todo: wipe

PR #143, opened 2016-04-20
Fix a bug that was returned the key nodes_callbacks that was removed

Ozahata:
- The changes was made in the commit 4a95342.
     First was removed in line 146 but had other change in 1024 that returned the key.

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -1020,7 +1020,6 @@ class RedisCluster(StrictRedisCluster):
             connection_pool=self.connection_pool,
             startup_nodes=self.connection_pool.nodes.startup_nodes,
             refresh_table_asap=self.refresh_table_asap,
-            nodes_callbacks=self.nodes_callbacks,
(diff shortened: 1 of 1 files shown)


Grokzen:
@Ozahata If this is the one you want me to merge, please close the other one and refer to this PR in that PR.

Grokzen:
@Ozahata you need to change the destination branch for this PR. The master branch is only used when i create a official release. All changes should me merged into the default branch named `unstable`. You can change the destination branch inside this PR without having to close this one and create a new one.

Grokzen:
@Ozahata I also need a description on what the problem is that this change will solve? Is there something that breaks when you run the code? Is this just a code cleanup or what will it affect if i merge this change? I do not get why we need this.

Ozahata:
@Grokzen I cannot change the destination branch in this PR (The edit button only allow me to edit the title) and all the answers that I see tell to close the PR and create a new one.
Since the PR #143 is the same, I will re-open and write the description there.

PR #144, opened 2016-04-21
Fix KeyError on node['name'] not in self._in_use_connections

ashishbaghudana:


Grokzen:
@ashishbaghudana I fixed the problem in the following commit 09145d59fea8cd9ef96a402dc6c9f0d6aa8a5f06 in a simpler way then a try/catch. I also added a line in Release notes file that you missed (with credeits to you for the fix) and i also added a test to ensure that it works in the future.

The one thing that i do not yet understand is in what scenario you managed to get this error to appear. The only scenario that i can see is that you tried to use the lib for a pubsub solution (even if i explicitly stated that it should not be used becuase of performance issues) and then you use the same client object instance in multiple threads and you happened to get a resharding event where the client rebuilt or was in the middle of rebuilding your node cache at the same time you tried to make a new pubsub instance in multiple threads. Do that sounds correct?

PR #154, opened 2016-08-06
Fix for Python 3 bytes key.  fix #153.

### LINKED ISSUE #153: NodeManager.keyslot() returns different numbers in Py2 and Py3
for example, give b'rq:job:bb8a8f67-1c83-454a-a170-891392d4d1ee' as its input,and RedisClusterHashSlots is 16384.
When running in Python 2.7.10, it returns 5963, while in Python3.5.1, it returns 14496.
This bug cause code to run into MoveError Exception at (n-1)/n probability when it write into a redis cluster.

fossilet:
It the key is bytes in Python 3, unicode(key) would be str(key), which is just an string representation of the bytes. We must provide encoding to convert it to an str type. I hope this fix the problem.

### DIFF
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -2,6 +2,7 @@
 
 # python std lib
 import random
+import sys
 
 # rediscluster imports
 from .crc import crc16
@@ -9,7 +10,7 @@ from .exceptions import RedisClusterException
 
 # 3rd party imports
 from redis import StrictRedis
-from redis._compat import unicode
+# from redis._compat import unicode
 from redis import ConnectionError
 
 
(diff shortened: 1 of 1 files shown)


Grokzen:
@fossilet Please add some tests that can validate your claims that it is broken between python versions and that then shows that this patch fixes the problem. 

Please read https://github.com/Grokzen/redis-py-cluster/blob/unstable/CONTRIBUTING.md#pull-request and include all required items in that section.

Grokzen:
@fossilet I manually took the changes and improved it to be mergable. You can view the commit here 3b4de9c84bcbab70a1557344c61c7cfde76fc06f Closing this issue as fixed with that commit.

Issue #156, opened 2016-09-07
the python cluster driver failed to failover sometimes


monklof:
## Problem

I created a redis cluster, and when I test the failover ability, sometimes the python driver can successfully failover after some time. But some times it failed to do so, it constantly raised ConnectionError.

```
ConnectionError: Error 111 connecting to 10.16.167.16:6379. Connection refused.
```

## Details: 

I have 6 nodes in three machine, A1 A2 A3 B1 B2 B3. An is the master node, Bn is the corresponding slave node. A1-3 run at port 6379, B1-3 run at port 6380. It is guaranteed that the master node and slave node are not in the same machine.

the master nodes' redis server conf: 

```
#redis config
daemonize yes
pidfile "/var/run/redis6379.pid"
port 6379
timeout 172800
tcp-keepalive 0
loglevel notice
logfile "/opt/redis6379/redis.log"   
databases 16

#persistence
#save 900 1
#save 300 10
#save 60 10000
stop-writes-on-bgsave-error yes
rdbcompression yes
rdbchecksum yes
dbfilename "redis6379.rdb"
dir "/opt/redis6379"
appendonly yes
appendfsync everysec
appendfilename "appendonly.aof"
no-appendfsync-on-rewrite yes

#replication
slave-serve-stale-data yes
slave-read-only yes
repl-ping-slave-period 10
repl-timeout 60
repl-disable-tcp-nodelay no
slave-priority 100

rename-command FLUSHALL FLUSHALL##   
rename-command FLUSHDB FLUSHDB##
rename-command KEYS KEYS##

maxclients 10000
maxmemory 7812500kb
maxmemory-policy noeviction
lua-time-limit 5000
slowlog-log-slower-than 10000
slowlog-max-len 1024
hash-max-ziplist-entries 512
hash-max-ziplist-value 64
list-max-ziplist-entries 512
list-max-ziplist-value 64
set-max-intset-entries 512
zset-max-ziplist-entries 128
zset-max-ziplist-value 64
activerehashing yes
client-output-buffer-limit normal 256mb 128mb 60
client-output-buffer-limit slave 512mb 256mb 180
client-output-buffer-limit pubsub 32mb 8mb 60
hz 10

#set bgrewriteaof
aof-rewrite-incremental-fsync yes
auto-aof-rewrite-percentage 1000
auto-aof-rewrite-min-size 64mb

#cluser
cluster-node-timeout 15000
cluster-config-file "nodes.conf"
[...]

PR #157, opened 2016-09-08
choose the currently discovered node to get cluster-require-full-cover config instead of the old ones

monklof:
#156


Grokzen:
@monklof Fix seems to be good. I manually merged it into unstable branch. You did not follow the contribution guide fully so i had to go in and add some things into this but the fix is in general small and works.

PR #160, opened 2016-09-17
better error handling around pipelined socket reads. …

72squared:
fixes a bug where a connection was being released into the pool with leftover response still left in the socket buffer, resulting in mangled or incorrect responses on subsequent queries.

I have been seeing really nasty errors whenever master nodes die and knew it was a result of pipeline socket connection logic but couldn't pin it down. the breakthrough came when I did a deliberate `CLUSTER FAILOVER` to a slave. Everything continued to work fine but when I killed off the old master (now a slave) all kinds of hell broke loose.

I spent a long time in a debugger in my test environment, setting breakpoints at various points and was able to repro the bug. The easiest way to repro the problem is to set a breakpoint in the old code on this line:

https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/pipeline.py#L207

When it hits the breakpoint, kill the server node it is about to write to. The write operation appears to succeed. But when it reads, an exception is thrown. If it was supposed to read from another socket after the first read, it won't get to it, but it will release it back into the connection pool, creating real problems for subsequent reads.

My patch attempts to fix and clean it up on several vectors. I never ever want to free a connection back into the pool if there is a chance it might not have completed the read after a write.

And if it does hit a read error, handle it more elegantly.


72squared:
Hold off on this pull request. I wrote some more test scenarios and found there are still problems. I want to spend the time to get this right.

72squared:
Okay this latest version should at least be functionally correct. I finally settled on simplifying pipeline logic a lot and only attempt to pipeline the commands the first time through. If I run into any errors, send the remaining commands that have not yet been sent one at a time through the normal logic. This makes it so that we use the more well tested and easier to understand logic for how to handle MOVED and ASKING errors, connection retry errors, timeouts etc. 
I've sent hours and hours running this script:

https://gist.github.com/72squared/92262949a90b6ea12e5375d68b6ff5e6

And then killing nodes in the cluster or failing over.

Grokzen:
@72squared It will probably take me some time to process through all of these changes. Probably until the weekend.

72squared:
No rush. I'll be giving it some bake time on our dev environment in the mean time.

rediscluster/pipeline.py
```diff
@@ -462,7 +397,23 @@ def read(self):
         """
         connection = self.connection
         for c in self.commands:
-            try:
-                c.result = self.parse_response(connection, c.args[0], **c.options)
-            except RedisError:
-                c.result = sys.exc_info()[1]
+
```
Grokzen: Remove blank line here

rediscluster/pipeline.py
```diff
@@ -142,144 +146,70 @@ def send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=T
         `allow_redirections` If the pipeline should follow `ASK` & `MOVED` responses
         automatically. If set to false it will raise RedisClusterException.
         """
-        if self.refresh_table_asap:
```
Grokzen: After removing this part the variable `self.refersh_table_asap` is not in use in the class so the variable is not needed in the `__init__` method. The refresh table logic is still in use in the regular client class.

rediscluster/pipeline.py
```diff
@@ -142,144 +146,70 @@ def send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=T
...
+        # next time we read from it we pass the buffered result back from a previous
+        # command and every single request after to that connection will always get
+        # a mismatched result. (not just theoretical, I saw this happen on production x.x).
+        for n in nodes.values():
+            self.connection_pool.release(n.connection)
+
+        # if the response isn't an exception it is a valid response from the node
+        # we're all done with that command, YAY!
+        # if we have more commands to attempt, we've run into problems.
+        attempt = sorted([c for c in attempt if isinstance(c.result, ERRORS_ALLOW_RETRY)], key=lambda x: x.position)
```
Grokzen: Can you think of a better name for this variable `attempt` because it do not feel like an attempt any longer as the logic for the attempts was moved to other places.

rediscluster/pipeline.py
```diff
@@ -1,40 +1,45 @@
...
-        self.refresh_table_asap = refresh_table_asap
-        self.reinitialize_counter = 0
-        self.reinitialize_steps = reinitialize_steps or 25
         self.response_callbacks = response_callbacks
         self.result_callbacks = result_callbacks
         self.startup_nodes = startup_nodes if startup_nodes else []
 
+        self.refresh_table_asap = False
+        self.nodes_flags = self.__class__.NODES_FLAGS.copy()
+        self.result_callbacks = self.__class__.RESULT_CALLBACKS.copy()
```
Grokzen: If we can send in result_callbacks and response_callbacks as vairables and then we just override it anyway with this line, would it not make sense to remove those 2 variables from `__init__` and just set them with this copy command?

rediscluster/pipeline.py
```diff
@@ -142,144 +146,70 @@ def send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=T
...
-                elif isinstance(c.result, Exception):
-                    break
-
-        # We are done trying. did we get stuck on connection or timeout errors?
-        if [c for c in attempt if isinstance(c.result, (ConnectionError, TimeoutError))] and ttl < 1:
-            raise ClusterError('TTL exhausted.')
-
-        # YAY! we made it out of the attempt loop.
+                try:
+                    c.result = super(StrictClusterPipeline, self).execute_command(*c.args, **c.options)
```
Grokzen: Can you add some comment like `This is where the magic ASK/MOVED redirection magic happens inside the execute_command that is piggybacking on the regular redirection logic from the normal client class` It was unclear to me at first where it was moved, but with a hint it is here it will make it much clearer.

rediscluster/pipeline.py
72squared: Well, we still do attempt the commands and if any fail, we attempt the remaining commands one at a time to retry them. I see your point, but haven't come up with anything better so far.

rediscluster/pipeline.py
72squared: 👍

rediscluster/pipeline.py
```diff
@@ -452,8 +382,13 @@ def write(self):
         connection = self.connection
         commands = self.commands
         try:
+            # We are going to clobber the commands with the write, so go ahead
+            # and ensure that nothing is sitting there from a previous run.
+            for c in commands:
```
72squared: 👍

rediscluster/pipeline.py
72squared: `StrictRedisCluster.execute_command` expects `refresh_table_asap` to be defined and blows up if it isn't.

rediscluster/pipeline.py
```diff
@@ -462,7 +397,23 @@ def read(self):
...
+            # result in an AttributeError because it will do a readline() call on None.
+            # This can have all kinds of nasty side-effects.
+            # Treating this case as a connection error is fine because it will dump
+            # the connection object back into the pool and on the next write, it will
+            # explicitly open the connection and all will be well.
+            if c.result is None:
+                try:
+                    c.result = self.parse_response(connection, c.args[0], **c.options)
+                except (ConnectionError, TimeoutError) as e:
+                    for c in self.commands:
```
72squared: The main block of pipeline `send_cluster_commands` will loop through all of these commands individually to see if we can retry them or not and we need the exception attached to each command so that it knows it can retry them.

Grokzen:
@72squared LGTM :) Merge time and release time.

Many thanks for doing the work and improve the situation for pipelines.

Issue #140, opened 2016-04-12
Improve support for same slot usage


Grokzen:
Currently the solution is designed around functions that has special implementation to support multi slot commands. What needs to be investigated and looked at is to support the more simple method of allowing multikey commands and pipelines to work when they are using the same slot for all keys. This would open up more functionality like WATCH and Transactions and enable the use of multi key functions to run on the server instead of inside the client.

The current proposal is to either have a config option that you send into `StrictRedis` or to have different classes that supports the different features.

AngusP:
Supporting hash tagged keys that have matching tags would be great. I think it'd be a bad idea to try and determine for *any* key if they're in the same slot that we should use the singular redis command (e.g. `BRPOPLPUSH`, set operation) as this would result in non-deterministic atomicity as clusters gain or loose nodes and reshard. 

The sample implementation for getting the slot of hashtagged keys [is given](https://redis.io/topics/cluster-spec#keys-hash-tags):

```ruby
def HASH_SLOT(key)
    s = key.index "{"
    if s
        e = key.index "}",s+1
        if e && e != s+1
            key = key[s+1..e-1]
        end
    end
    crc16(key) % 16384
end
```

Having detection for safe atomic operations would likely give significant speedups in pipelines, large sets and other places, as well as give clusters the same atomicity guarantees as singular Redis servers.

Perhaps a way of enforcing safety would also be useful, such as:

```python
r = rediscluster.StrictRedisCluster(...)
with r.guarantee_sameslot():
    p = r.pipeline()
    p.lpush(_, _)
    ...
    p.execute()
```

Or some flag that can be passed to methods with custom cluster implementations asking them to raise exceptions if they can't use the redis command instead of falling back on a cluster impl.

PR #162, opened 2016-10-13
Support for multi key commands MGET & MSET to pipelines

Grokzen:
First itteration to enable support for MGET and MSET when used to same hash slot when used in pipelines.


Grokzen:
Will be a step to solve #140

PR #164, opened 2016-10-19
Added options to skip full coverage check

dutradda:
I wrote this patch because the AWS Elasticache disables the CONFIG command. Recently aws launched the redis 3.2 engine.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -20,3 +20,4 @@ Authors who contributed code or testing:
  - gmolight - https://github.com/gmolight
  - baranbartu - https://github.com/baranbartu
  - monklof - https://github.com/monklof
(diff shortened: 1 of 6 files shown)


Grokzen:
@dutradda Looks good from a a quick look at it. A few things that you need to add to this. First is a line in ReleaseNotes file mentioning the addition of this new feature/flag. The second is to add yourself to Authors file if you want. The third is to look if there is any good spot to put in some docs about this flag/feature. I can't think of one right away but look to see if there is one.

dutradda:
@Grokzen I made the changes that you ask for

Grokzen:
@dutradda Thanks for this. LGTM.

PR #165, opened 2016-10-31
Handle error when connecting to Elasticache

cnu:
Elasticache has CONFIG commands disabled.
So config_get to get if the cluster has full coverage won't work.
Handle the error and return False as default.


Grokzen:
@cnu #164 already solves this in a better way. Will not merge this over that one.

PR #170, opened 2016-12-09
Add support for bytes keys that are not utf-8 compatible

0xabe-io:
The change enables the support of all possible bytes object as the redis-py library does. As of now if bytes that are not available in utf-8 are used in keys, the client crashes:
```
>>> con = rediscluster.StrictRedisCluster(startup_nodes=NODES)
[{'port': '6379', 'host': '213.244.194.42'}, {'port': '6379', 'host': '213.244.194.43'}, {'port': '6379', 'host': '213.244.194.44'}, {'port': '6379', 'host': '213.244.194.70'}, {'port': '6379', 'host': '213.244.194.78'}, {'port': '6379', 'host': '213.244.194.81'}]
>>> con.set(b'\x01', 0x01)
True
>>> con.set(b'\xa1', 0xa1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/site-packages/redis/client.py", line 1072, in set
    return self.execute_command('SET', *pieces)
  File "/home/abe/research/libcmonger/rediscluster/utils.py", line 101, in inner
    return func(*args, **kwargs)
  File "/home/abe/research/libcmonger/rediscluster/client.py", line 308, in execute_command
    slot = self._determine_slot(*args)
  File "/home/abe/research/libcmonger/rediscluster/client.py", line 254, in _determine_slot
    return self.connection_pool.nodes.keyslot(key)
  File "/home/abe/research/libcmonger/rediscluster/nodemanager.py", line 70, in keyslot_py_3
    k = str(key, encoding='utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa1 in position 0: invalid start byte
```

Instead of converting a bytes back to str, just use bytes for all keys as redis-py does:
```
>>> con.set(b'\xa1', 0xa1)
True
>>> hex(int(con.get(b'\xa1')))
'0xa1'
```

### DIFF
--- a/rediscluster/crc.py
+++ b/rediscluster/crc.py
@@ -43,7 +43,9 @@ def _crc16_py3(data):
     """
     """
     crc = 0
-    for byte in data.encode("utf-8"):
+    if hasattr(data, "encode"):
+        data = data.encode("utf-8")
+    for byte in data:
         crc = ((crc << 8) & 0xff00) ^ x_mode_m_crc16_lookup[((crc >> 8) & 0xff) ^ byte]
     return crc & 0xffff
 
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -67,8 +67,10 @@ class NodeManager(object):
         """
         if hasattr(key, "encode"):
             k = key.encode("utf-8")
-        else:
+        elif hasattr(key, "decode"):
             k = key
+        else:
+            k = str(key).encode()
 
         start = k.find(b"{")
 
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -65,17 +65,15 @@ class NodeManager(object):
         Calculate keyslot for a given key.
         Tuned for compatibility with supported python 3.x versions
         """
-        try:
-            # Handle bytes case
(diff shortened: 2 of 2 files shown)


Grokzen:
@0xabe-io Looks good from initial review. One thing is that if you could add some minor tests for this method that shows/tests the different cases and let travis test them in the different python versions to see they are all compatible? https://github.com/Grokzen/redis-py-cluster/blob/unstable/tests/test_node_manager.py#L32 is a good place to add more values that should be supported.

AngusP:
If you're having to fiddle with encoding and decoding, `redis-py` provides a nice method that'll work with both Python 2 and Python 3 called `nativestr` which is in `_compat.py`

Grokzen:
@0xabe-io Bump, please see my previous comments. Would like to get this merged :)

Grokzen:
@0xabe-io @AngusP Please see the reference MR #186 for a better fix. Closing this one becuase #186 superseeds this one with even more features and more up to date.

If you find any problems with the impl in #186 please open up a new issue. 

Closing this one as "won't merge"

Issue #171, opened 2016-12-09
`cluster_nodes` throwing exception


AngusP:
Calling the `rediscluser.StrictRedisCluster` method `cluster_nodes()` raises an `AttributeError` if the class has not been initialised with the `decode_responses` argument (Which should be false and defaults to false if one wants to store binary data in Redis)

To reproduce, using Python 3.4.2
```python
In [1]: import rediscluster

In [2]: r = rediscluster.StrictRedisCluster(startup_nodes=[{'host':'127.0.0.1','port':'6379'}] )
[{'host': '127.0.0.1', 'port': '6379'}]

In [3]: r.cluster_nodes()
'''
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-af3a2ae28b27> in <module>()
----> 1 r.cluster_nodes()

/wherever/VENV/lib64/python3.4/site-packages/rediscluster/client.py in cluster_nodes(self)
    488         Sends to random node in the cluster
    489         """
--> 490         return self.execute_command('CLUSTER NODES')
    491 
    492     def cluster_replicate(self, target_node_id):

/wherever/VENV/lib64/python3.4/site-packages/rediscluster/utils.py in inner(*args, **kwargs)
     99         for _ in range(0, 3):
    100             try:
--> 101                 return func(*args, **kwargs)
    102             except ClusterDownError:
    103                 # Try again with the new cluster setup. All other errors

/wherever/VENV/lib64/python3.4/site-packages/rediscluster/client.py in execute_command(self, *args, **kwargs)
    295         node = self.determine_node(*args, **kwargs)
    296         if node:
--> 297             return self._execute_command_on_nodes(node, *args, **kwargs)
    298 
    299         # If set externally we must update it before calling any commands

/wherever/VENV/lib64/python3.4/site-packages/rediscluster/client.py in _execute_command_on_nodes(self, nodes, *args, **kwargs)
    379             try:
    380                 connection.send_command(*args)
--> 381                 res[node["name"]] = self.parse_response(connection, command, **kwargs)
    382             except (ConnectionError, TimeoutError) as e:
    383                 connection.disconnect()

/wherever/VENV/lib64/python3.4/site-packages/redis/client.py in parse_response(self, connection, command_name, **options)
    585         response = connection.read_response()
    586         if command_name in self.response_callbacks:
--> 587             return self.response_callbacks[command_name](response, **options)
    588         return response
    589 

[...]

PR #175, opened 2017-01-01
Convert raw responses from Redis to native strings in CLUSTER command parsing, allowing the use of decode_responses=False in Python 3+

AngusP:
Using `nativestr` from `redis._compat` to convert responses passed to the parsing callbacks `parse_cluster_nodes` and `parse_cluster_slots` in `rediscluster/utils.py`. 

This addresses issue #171 where undecoded responses in Python 3 are of type `bytes`, which iterates per-byte as opposed to per-line like `str`, breaking the parsing method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,6 @@ Small sample script that shows how to get started with RedisCluster. It can also
 >>> # Requires at least one node for cluster discovery. Multiple nodes is recommended.
 >>> startup_nodes = [{"host": "127.0.0.1", "port": "7000"}]
 
->>> # Note: decode_responses must be set to True when used with python3
 >>> rc = StrictRedisCluster(startup_nodes=startup_nodes, decode_responses=True)
 
 >>> rc.set("foo", "bar")
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -44,7 +44,7 @@ Small sample script that shows how to get started with RedisCluster. It can also
     >>> # Requires at least one node for cluster discovery. Multiple nodes is recommended.
     >>> startup_nodes = [{"host": "127.0.0.1", "port": "7000"}]
 
-    >>> # Note: decode_responses must be set to True when used with python3
+    >>> # Note: decode_responses behaves differently with python3 and python2
     >>> rc = StrictRedisCluster(startup_nodes=startup_nodes, decode_responses=True)
 
     >>> rc.set("foo", "bar")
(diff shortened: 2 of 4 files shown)


AngusP:
@Grokzen have you had a chance to look this over?

docs/index.rst
```diff
@@ -44,15 +44,20 @@ Small sample script that shows how to get started with RedisCluster. It can also
     >>> # Requires at least one node for cluster discovery. Multiple nodes is recommended.
     >>> startup_nodes = [{"host": "127.0.0.1", "port": "7000"}]
 
-    >>> # Note: decode_responses must be set to True when used with python3
+    >>> # Note: decode_responses gives different results with python3 and python2
```
Grokzen: I agree that removing the `must be set to True...` is good to remove, but i am not sold on the change to `gives different results...` that is very ambiguous and non descriptive. Suggestion is to tell something like `When set to True in python3 it will convert all binary responses into strings` or something similar.

docs/index.rst
```diff
@@ -44,15 +44,20 @@ Small sample script that shows how to get started with RedisCluster. It can also
...
     >>> rc = StrictRedisCluster(startup_nodes=startup_nodes, decode_responses=True)
 
     >>> rc.set("foo", "bar")
     True
     >>> print(rc.get("foo"))
     'bar'
 
+It is fairly safe to assume that any methods not documented here wil behave in the same way as
```
Grokzen: Not sold on the working `...failry safe...`. Please give it some more consideration.

rediscluster/utils.py
```diff
@@ -126,7 +126,7 @@ def parse_cluster_slots(resp, **options):
     current_host = options.get('current_host', '')
 
     def fix_server(*args):
-        return (args[0] or current_host, args[1])
+        return (nativestr(args[0]) or current_host, args[1])
```
Grokzen: I would love to see a testcase in this method `test_parse_cluster_slots` that tests the same data structure but with the data format that you get back from redis when using `parse_responses=False` and without this fix. It would show that both dataformats would work as expected.

README.md
```diff
@@ -59,4 +58,4 @@ Copyright (c) 2013-2017 Johan Andersson
 
 MIT (See docs/License.txt file)
 
-The license should be the same as redis-py (https://github.com/andymccurdy/redis-py)
\ No newline at end of file
+The license should be the same as redis-py (https://github.com/andymccurdy/redis-py)
```
Grokzen: I do not understand what github says that was changed on this line. Do you know? It do not show any content change on the line and no change to end of file O.o really wierd.

docs/index.rst
AngusP: How about

> When set to False, redis-py-cluster will not attempt to decode the responses it receives. In Python 3, this means the responses will be of type `bytes`. In Python 2, they will be native strings (`str`). If `decode_responses` is set to True, for Python 3 responses will be `str`, for Python 2 they will be `unicode`.

docs/index.rst
AngusP: Probably should just be removed, given a lot of set and list methods are overridden and no longer atomic

rediscluster/utils.py
AngusP: Will do

README.md
AngusP: It might have been that the file had no trailing newline?

AngusP:
@Grokzen I've pushed changes based on review, my comments on the review reflect what has changed.

AngusP:
@Grokzen Addressed review in latest commit 9d32674

PR #181, opened 2017-01-25
Add PUBSUB * commands, sending to all nodes

AngusP:
PUBSUB commands added to redis-py in commit [3f4ac6](https://github.com/andymccurdy/redis-py/commit/3f4ac6e685b122f49b516db3f59c1edc988e6b49), which should be sent to all nodes (https://redis.io/commands/pubsub, https://redis.io/topics/pubsub)

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -50,14 +50,14 @@ class StrictRedisCluster(StrictRedis):
             "ECHO", "CONFIG GET", "CONFIG SET", "SLOWLOG GET", "CLIENT KILL", "INFO",
             "BGREWRITEAOF", "BGSAVE", "CLIENT LIST", "CLIENT GETNAME", "CONFIG RESETSTAT",
             "CONFIG REWRITE", "DBSIZE", "LASTSAVE", "PING", "SAVE", "SLOWLOG LEN", "SLOWLOG RESET",
-            "TIME", "KEYS", "CLUSTER INFO",
+            "TIME", "KEYS", "CLUSTER INFO", "PUBSUB CHANNELS",  
+            "PUBSUB NUMSUB", "PUBSUB NUMPAT",
         ], 'all-nodes'),
         string_keys_to_dict([
             "FLUSHALL", "FLUSHDB", "SCRIPT LOAD", "SCRIPT FLUSH", "SCRIPT EXISTS", "SCAN",
         ], 'all-masters'),
         string_keys_to_dict([
-            "RANDOMKEY", "CLUSTER NODES", "CLUSTER SLOTS", "PUBSUB CHANNELS",  
-            "PUBSUB NUMSUB", "PUBSUB NUMPAT",
+            "RANDOMKEY", "CLUSTER NODES", "CLUSTER SLOTS",
         ], 'random'),
         string_keys_to_dict([
             "CLUSTER COUNTKEYSINSLOT",
@@ -71,6 +71,7 @@ class StrictRedisCluster(StrictRedis):
             "CONFIG REWRITE", "DBSIZE", "LASTSAVE", "PING", "SAVE", "SLOWLOG LEN", "SLOWLOG RESET",
             "TIME", "SCAN", "CLUSTER INFO", 'CLUSTER ADDSLOTS', 'CLUSTER COUNT-FAILURE-REPORTS',
             'CLUSTER DELSLOTS', 'CLUSTER FAILOVER', 'CLUSTER FORGET', "FLUSHALL", "FLUSHDB",
+            "PUBSUB CHANNELS", "PUBSUB NUMSUB", "PUBSUB NUMPAT",
         ], lambda command, res: res),
         string_keys_to_dict([
             "SCRIPT LOAD",
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -56,7 +56,8 @@ class StrictRedisCluster(StrictRedis):
             "FLUSHALL", "FLUSHDB", "SCRIPT LOAD", "SCRIPT FLUSH", "SCRIPT EXISTS", "SCAN",
         ], 'all-masters'),
         string_keys_to_dict([
-            "RANDOMKEY", "CLUSTER NODES", 'CLUSTER SLOTS',
+            "RANDOMKEY", "CLUSTER NODES", "CLUSTER SLOTS", "PUBSUB CHANNELS",  
+            "PUBSUB NUMSUB", "PUBSUB NUMPAT",
         ], 'random'),
         string_keys_to_dict([
             "CLUSTER COUNTKEYSINSLOT",
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -91,10 +91,10 @@ class StrictRedisCluster(StrictRedis):
             "SSCAN", "HSCAN", "ZSCAN", "RANDOMKEY",
         ], first_key),
         {
-            "PUBSUB CHANNELS" : parse_pubsub_channels,
-            "PUBSUB NUMSUB"   : parse_pubsub_numsub,
-            "PUBSUB NUMPAT"   : parse_pubsub_numpat,
-        },
+            "PUBSUB CHANNELS": parse_pubsub_channels,
+            "PUBSUB NUMSUB": parse_pubsub_numsub,
+            "PUBSUB NUMPAT": parse_pubsub_numpat,
+        }
     )
 
     CLUSTER_COMMANDS_RESPONSE_CALLBACKS = {
@@ -262,12 +262,12 @@ class StrictRedisCluster(StrictRedis):
 
         return self.connection_pool.nodes.keyslot(key)
 
-    def _merge_result(self, command, res):
+    def _merge_result(self, command, res, **kwargs):
         """
         `res` is a dict with the following structure Dict(NodeName, CommandResult)
         """
         if command in self.result_callbacks:
-            return self.result_callbacks[command](command, res)
+            return self.result_callbacks[command](command, res, **kwargs)
 
         # Default way to handle result
         return first_key(command, res)
@@ -399,7 +399,7 @@ class StrictRedisCluster(StrictRedis):
             finally:
                 self.connection_pool.release(connection)
 
-        return self._merge_result(command, res)
+        return self._merge_result(command, res, **kwargs)
 
     ##########
     # Cluster management commands
@@ -751,8 +751,7 @@ class StrictRedisCluster(StrictRedis):
         Return a list of channels that have at least one subscriber.
         Aggregate toggles merging of response.
         """
-        options = { 'aggregate': aggregate }
-        return self.execute_command('PUBSUB CHANNELS', pattern, **options)
+        return self.execute_command('PUBSUB CHANNELS', pattern, aggregate=aggregate)
 
 
     def pubsub_numpat(self, aggregate=True):
@@ -760,17 +759,17 @@ class StrictRedisCluster(StrictRedis):
         Returns the number of subscriptions to patterns.
         Aggregate toggles merging of response.
         """
-        options = { 'aggregate': aggregate }
-        return self.execute_command('PUBSUB NUMPAT', **options)
+        return self.execute_command('PUBSUB NUMPAT', aggregate=aggregate)
 
 
-    def pubsub_numsub(self, *args, aggregate=True):
+    def pubsub_numsub(self, *args, **kwargs):
         """
         Return a list of (channel, number of subscribers) tuples
         for each channel given in ``*args``.
-        Aggregate toggles merging of response.
+        
+        ``aggregate`` keyword argument toggles merging of response.
         """
-        options = { 'aggregate': aggregate }
+        options = { 'aggregate': kwargs.get('aggregate', True) }
         return self.execute_command('PUBSUB NUMSUB', *args, **options)
 
     ####
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -23,6 +23,9 @@ from .utils import (
     clusterdown_wrapper,
     parse_cluster_slots,
     parse_cluster_nodes,
+    parse_pubsub_channels,
+    parse_pubsub_numsub,
+    parse_pubsub_numpat,
 )
 # 3rd party imports
 from redis import StrictRedis
@@ -71,7 +74,6 @@ class StrictRedisCluster(StrictRedis):
             "CONFIG REWRITE", "DBSIZE", "LASTSAVE", "PING", "SAVE", "SLOWLOG LEN", "SLOWLOG RESET",
             "TIME", "SCAN", "CLUSTER INFO", 'CLUSTER ADDSLOTS', 'CLUSTER COUNT-FAILURE-REPORTS',
             'CLUSTER DELSLOTS', 'CLUSTER FAILOVER', 'CLUSTER FORGET', "FLUSHALL", "FLUSHDB",
-            "PUBSUB CHANNELS", "PUBSUB NUMSUB", "PUBSUB NUMPAT",
         ], lambda command, res: res),
         string_keys_to_dict([
             "SCRIPT LOAD",
@@ -88,6 +90,11 @@ class StrictRedisCluster(StrictRedis):
         string_keys_to_dict([
             "SSCAN", "HSCAN", "ZSCAN", "RANDOMKEY",
         ], first_key),
+        {
+            "PUBSUB CHANNELS" : parse_pubsub_channels,
+            "PUBSUB NUMSUB"   : parse_pubsub_numsub,
+            "PUBSUB NUMPAT"   : parse_pubsub_numpat,
+        },
     )
 
     CLUSTER_COMMANDS_RESPONSE_CALLBACKS = {
@@ -738,6 +745,34 @@ class StrictRedisCluster(StrictRedis):
 
         return False
 
+
+    def pubsub_channels(self, pattern='*', aggregate=True):
+        """
+        Return a list of channels that have at least one subscriber.
+        Aggregate toggles merging of response.
+        """
+        options = { 'aggregate': aggregate }
+        return self.execute_command('PUBSUB CHANNELS', pattern, **options)
+
+
+    def pubsub_numpat(self, aggregate=True):
+        """
+        Returns the number of subscriptions to patterns.
+        Aggregate toggles merging of response.
+        """
+        options = { 'aggregate': aggregate }
+        return self.execute_command('PUBSUB NUMPAT', **options)
+
+
+    def pubsub_numsub(self, *args, aggregate=True):
+        """
+        Return a list of (channel, number of subscribers) tuples
+        for each channel given in ``*args``.
+        Aggregate toggles merging of response.
+        """
+        options = { 'aggregate': aggregate }
+        return self.execute_command('PUBSUB NUMSUB', *args, **options)
+
     ####
     # List commands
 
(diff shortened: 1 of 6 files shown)


Grokzen:
@AngusP I do not have any issues with the impl, it looks good and no problems there. The thing that i do not really know how to handle in this specific case is that since redis-py have not made a release with this change and i do not see a release in the near future atm, i am not sure when to merge this really. I can merge it now but then everyone that want to use it has to wait for the release before it can be used, or you have to use the develop branch from redis-py.

On the other hand it might not hurt anything anyway since it is only enabling it to route to a random node and it do not include any special method that overwrites a method in redis-py.

AngusP:
@Grokzen the pull into redis-py addressed a pretty old [issue](https://github.com/andymccurdy/redis-py/issues/526) from 2014, these commands have been in Redis for a while. Merging this would at least let `execute_command('PUBSUB CHANNELS')` work without crashing out, and when it makes it into a redis-py release, redis-py-cluster will already work with the new commands

Grokzen:
The one thing that i think about is that becuase pubsub is a bit whacky implemented inside redis itself it cause problems when used in a cluster setting.

The main problem is that you can have clients listen to the same channel on any of the nodes in the cluster and messages is passed to each listener properly. But the problem is that when redis responds back to the client about how many that received the message it only reports back how many clients that returned from the same node.

So how this related to your impl is that if you choose to set it into a random node it might have the same problem that redis itself returns wrong data in certain cases depending on how to works internally with tracking and reporting the data.

Unless tested, it might be that redis only counts the clients connected to the same node as you query and not counts the total number across the entire cluster. Unless you can't show that this is the case, i do think that random is a bad choice as it can cause problems down the line that is not shown in simple tests but fails in bigger scenarios.

AngusP:
@Grokzen I'll investigate sending to all nodes, and set up a bunch of subscribers, to see whether sending to one node randomly is safe. From initial testing you seem to be correct, at least `pubsub_channels` only lists those for which the node has at least one subscriber.

AngusP:
@Grokzen tested with all 3 sent to all nodes, Redis does indeed only return stats/channels for the node you send the command to

AngusP:
(As a note, my test environment was 3 masters, 3 slaves with 1 channel, 1 publisher and 4 subscribers arranged 1,2,0,0,0,1 over the cluster)

Grokzen:
@AngusP No problems :) There is already a pubsub implementation that kinda works by using a bad model just to make it appear to work but in reality it just cost more to have a client that "looks" like it is cluster enabled when in reality you pay a extreme high cost with 0 gains.

I think i have stated in the docs that using pubsub in a cluster is not recommended because of the reasons above. Sure you can make it work in your case but i would recommend that you build a solution with sentinel or manual handling of multiple nodes that do not pay the cost of the poor cluster impl so you acctually gain something out of it.

AngusP:
@Grokzen could you clarify whether that means you will merge or not? While I get that clusterised Redis' pubsub isn't great, it *is* still a feature of Redis so for completeness' sake and for those that don't particularly care about the speed costs it's useful to have

Grokzen:
Yes i do think i can merge the feature even if it is kinda broken just for completeness sake.

But i think that the implementation should be altered a little bit.

If you look over at the regular pubsub impelmentation here https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/pubsub.py#L26 and continue to track down to here https://github.com/Grokzen/redis-py-cluster/blob/956bf96456c37883a7ae89bc0c5bdbdc4b96ef96/rediscluster/connection.py#L157 you will find out that the strategy i have taken right now is that every pubsub subscription that is _not_ a pattern subscription will hash the channel name and get the keyslot and use that to determine to what node it should talk to. This will solve the problem when not using patterns so that all clients will go to the same node and the commands might work out a little bit better then if you send the command to a random node.

The problem with this approach is that it is not standard and implemented in all clients and preferred way of solving this from a redis pov. But this is how i choosen to do it and i do think that this is the way that redis will implement it when they actually fix and optimize pubsub in redis.

So i do think that you should change it from random to use the regular key hashing solution and for that i do think that you only need to remove the command from the random key and just let it go through `execute_command('PUBSUB CHANNELS')` by the default implementation that already do the key hashing functionality.

Grokzen:
So if i remember correctly on how the default impl works with slots hashing, the solution here might be to not do anything as that is the correct solution to make it compliant with the other PUBSUB impl :D

AngusP:
> But this is how i choosen to do it and i do think that this is the way that redis will implement it when they actually fix and optimize pubsub in redis.

Perhaps, although I've seen Bloom filters mentioned, though not exactly *how* they'd be used. I doubt PubSub will be optimised until Redis Cluster sees more widespread use, which is a shame.

> The problem with this approach is that it is not standard and implemented in all clients and preferred way of solving this from a redis pov. But this is how i choosen to do it and i do think that this is the way that redis will implement it when they actually fix and optimize pubsub in redis.

This would give counter-intuitive results if clients other than `redis-py-cluster` instances where talking to Redis, as an incomplete but nonempty set of channels would be returned by `pubsub_channels()`.

For my use case, it is also more valuable to be able to inspect the state of pubsub across the whole cluster than just the nodes who have the hash slots that channels' names fall into.

> So if i remember correctly on how the default impl works with slots hashing, the solution here might be to not do anything as that is the correct solution to make it compliant with the other PUBSUB impl :D

If by "not do anything" you mean don't add a member to `NODES_FLAGS`, a `RedisClusterException` _"No way to dispatch this command to Redis Cluster. Missing key."_ gets raised, which is what surprised me when I used the dev version of Redis-Py

Grokzen:
Right, forgot how it worked...

What happens is that here we take the RESPONSE_CALLBACKS that is defined in redis-py and merge it with the additional callbacks that needs to be defined in a cluster env. https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/client.py#L172 and they are merged with these ones https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L356

So in a normal case the callbacks is already defined in redis-py and that is considered a implmeneted method and i do think that then it would "just work" because it would then be present in that datastructure in the cluster client instance and it would understand that one.

How ever becuase the commands is arguments less there is no key to hash against and that is why you see the `no way to dispatch...` error because it comes from this check https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/client.py#L240

I would recommend that for your usecase where you want to execute it in a different manner then originally intended for the method to itterate over all master nodes on your end by looping over the output from this method call https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/client.py#L277 and do something like it is already done here https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/client.py#L297 in order to execute the command in the way you want. By doing it this way you should be able to get around the blocking restriction becuase after the block there should be nothing that stops you really from executing the command directly to each master node in the cluster.

Yes i do know that it is a problem with other clients not doing the very same thing, but that is because the reference implementation do not define a way to solve the problem, so i choose the best solution to make it work as much as possible and to make it work as equal as possible when moving from a non-cluster to a cluster client.

Yes i have seen the bloom filter thingy and that might be the solution forward to fix the performance problems inside redis but i have seen no progress on them so far. I can only hope, but if more people pressure for that to be fixed it might get higher priority, who knows...

For the utility use-case i would say that it is not a good way to solve it to add it for example to this one `all-nodes` becuase that would distort the expected return value from the method. It would return a dict with each node and the result and you can parse it however you want, but the core method should work similar to how it is expected to work in a non-cluster case.

We used to have this helper class before to enable some inspection capability and returning of data from the cluster but i removed it because it was not the right place to have custom solutions that is not standard in a way.

AngusP:
As I see it, the `PUBSUB CHANNELS`and `PUBSUB NUMPAT` commands are similar to `INFO` in that they should be used infrequently and have little or no use to clients in a production setting. Therefore being slow isn't much of a concern. `PUBSUB NUMSUB [argument [argument...]]` has more uses for a large production app, say counting the number of end users in live chat, whatever.

I'd be very wary of reinterpreting the behaviours of commands and behaving in an inconsistent way to how `redis-cli` or `redis-rb` do. Introducing quirks isn't a good idea in my opinion. Yes, I could use lower level bindings within redis-py and redis-py-cluster though I don't feel I should have to, if that makes sense? These are, after all, fairly simple commands understood by most Redis servers. (Though seemingly rarely used, judging by the small amount of attention their lack of inclusion in redi-py got.)

> I would recommend that for your usecase where you want to execute it in a different manner than originally intended

I don't really get what you mean, as the original intention is for the commands to receive a response per-server

AngusP:
> I might think of doing a dual solution so that the regular command querys all nodes and just counts the values and add them all together and return the same data as in a non-cluster, and then i either add a flag to the method or make a separate method like pubsub_numsub_all()

@Grokzen I would agree, having a 'normal' behaving and clusterised set of commands would make sense, probably via the second method option. I'd be happy to work towards that

Grokzen:
@AngusP Sounds good, make a suggestion implementation and ping me when you have something to review.

AngusP:
@Grokzen would you mind if I included a couple of minor bugfixes in this PR, or would you prefer a new one? For instance, the format string in the method `StrictRedisCluster._random_good_hashslot_key` is not doing what it should be:

**Existing:**
```python
random_id = "{{0}}{1}".format(hashslot, self._random_id())
>>> r._random_good_hashslot_key(10)
>>> '{0}GFGTLAMH8EIRCRGI'
```

**Correct:**
```python
random_id = "{{{0}}}{1}".format(hashslot, self._random_id())
>>> r._random_good_hashslot_key(10)
>>> '{10}GFGTLAMH8EIRCRGI'
```

AngusP:
@Grokzen I've got an initial interface that seems to be in-line with what we discussed, emulating the `redis-py` behaviour and returning the same types. I've effectively overridden the `redis-py` methods, adding a keyword argument `aggregate` defaulted to `True`.

I'm having trouble getting the `**options` passed from execute_command -> parse_response -> callbacks to actually be passed to the callback, so as it stands `aggregate` is effectively ignored.

There's also a slight quirk, wherein `PUBSUB NUMSUB [channel] [channels...]` won't return the `(channel, numsub)` list in the same order as they were called in `NUMSUB`, which is what Redis does. It doesn't seem to be 'promised' that this is the order they'll be in in the list, but both `redis-cli` and `redis-py` have them in this order. The order changes because I use a dict to sum the subscriber counts for all nodes in `utils.py/parse_pubsub_numpub`. This could be changed though the dict method will of course be faster than any list search for cases when there are a very large number of channels.

AngusP:
```python
def pubsub_numsub(self, *args, aggregate=True):
```
I'll also fix this bad function declaration

rediscluster/client.py
```diff
@@ -736,6 +745,33 @@ def renamenx(self, src, dst):
 
         return False
 
+
+    def pubsub_channels(self, pattern='*', aggregate=True):
```
Grokzen: I am not convinced that in the method `pubsub_channels` you have the defined keyword arguments `pattern` and `aggregate` and no `*args and **kwargs`. But in the method `pubsub_numsub` you have both and they all use/have the option to input `aggregate` as a kwarg. Please be consistent to use either or solution.

rediscluster/utils.py
```diff
@@ -203,3 +203,53 @@ def parse_slots(s):
         nodes.append(node)
 
     return nodes
+
+
+def parse_pubsub_channels(command, res, **options):
```
Grokzen: Why have this in utils.py? Why not have this code inside each method in the client class? These methods is not reused elsewhere but in the specific methods.

rediscluster/client.py
```diff
@@ -86,6 +90,11 @@ class StrictRedisCluster(StrictRedis):
         string_keys_to_dict([
             "SSCAN", "HSCAN", "ZSCAN", "RANDOMKEY",
         ], first_key),
+        {
```
Grokzen: Please be consitent with all other defentiions in this variable. Even if it is only one key to method you should use `string_keys_to_dict` to define all entries in RESULT_CALLBACKS

AngusP:
@Grokzen Build errors are due to the `PUBSUB` commands not yet being in a released version of `redis-py` (Added in [PR819](https://github.com/andymccurdy/redis-py/pull/819)) , so the result isn't correctly lumped into tuples.

**Latest:**
```python
>>> r.pubsub_numsub('foo','bar','baz')
[(b'foo', 0), (b'bar', 0), (b'baz', 0)]
```
**Released:**
```
>>> r.execute_command('PUBSUB','NUMSUB','foo','bar','baz')
[b'foo', 0, b'bar', 0, b'baz', 0
```

rediscluster/client.py
```diff
@@ -1104,7 +1140,7 @@ def _random_good_hashslot_key(self, hashslot):
         Generate a good random key with a low probability of collision between any other key.
         """
         # TODO: Check if the key exists or not. continue to randomize until a empty key is found
-        random_id = "{{0}}{1}".format(hashslot, self._random_id())
+        random_id = "{{{0}}}{1}".format(hashslot, self._random_id())
```
Grokzen: What problem do this solve? Will this not break the output from this method when x2 { is kept in the output string?

tests/test_pubsub.py
```diff
@@ -474,3 +475,31 @@ def t_run(rc):
...
+    @skip_if_server_version_lt('2.8.0')
+    def test_pubsub_channels(self, r):
+        p = r.pubsub(ignore_subscribe_messages=True)
+        p.subscribe('foo', 'bar', 'baz', 'quux')
+        channels = sorted(r.pubsub_channels())
+        assert channels == [b('bar'), b('baz'), b('foo'), b('quux')]
+
+    @skip_if_server_version_lt('2.8.0')
+    def test_pubsub_numsub(self, r):
+        p1 = r.pubsub(ignore_subscribe_messages=True)
```
Grokzen: could this not be optimized to only `r.pubsub(ignore_subscribe_messages=True).subscribe('foo', 'bar', baz')` No need for the variabel defenition then.

tests/test_pubsub.py
```diff
@@ -474,3 +475,31 @@ def t_run(rc):
             t.start()
     except Exception:
         print("Error: unable to start thread")
+
+
+class TestPubSubPubSubSubcommands(object):
+
+    @skip_if_server_version_lt('2.8.0')
```
Grokzen: In one way this blocking is kinda unnessesary becuase this lib will never test against any redis installation that is less then 3.0 becuase that was the version cluster support was introduced. I will add to my todo to remove it elsewhere later but no need to have it here.

tests/test_pubsub.py
```diff
@@ -474,3 +475,31 @@ def t_run(rc):
             t.start()
     except Exception:
         print("Error: unable to start thread")
+
+
+class TestPubSubPubSubSubcommands(object):
```
Grokzen: I do not really get the hang of the class name in realtion to what is being tested here O.o Either change to something that i can understand or add a docstring that describeres the purpose of this seperate class instead of having the tests closer to the regular pubsub tests.

Grokzen:
@AngusP Also saw that you are missing to add a line in the release notes for this change.

Grokzen:
@AngusP I am not really sure how we are going to tackle the problem with different code behaviour. As it stands right now i only support the code base that is released and if that code is not realsed we either have to not merge this until that time or we have to work around the problem.

To work around it right now the only way i see is that you check the internal data type of the objects and if it is not a tuple then you have to reconstruct the list into the proper form.

AngusP:
@Grokzen thanks for reviewing, will go through and comment individually re. changes. Also for your most recent comment, I shall add `@skip_if_redis_py_version_lt('2.10.5')` to tests that require redis-py to support the PUBSUB commands - any version earlier than that the method won't work without a custom response callback. In the long term, this won't be a problem, just the short term while one releases a feature the other doesn't have yet.

> To work around it right now the only way i see is that you check the internal data type of the objects and if it is not a tuple then you have to reconstruct the list into the proper form.

imho a work around is inelegant, and as I mentioned not needed long-term

rediscluster/client.py
AngusP: Will do

Grokzen:
@AngusP Agree, that solution with that form of skip will work very well.

rediscluster/client.py
AngusP: I went with the named arguments where possible, which allows tools like iPython to 'know' what the arguments are.

For `pubsub_channels`, `def pubsub_channels(self, *args, aggregate=True)` would work in Python3 and not Python2. I could use `*args, **kwargs` for all three but that looses the self-documenting of named keyword arguments.

rediscluster/client.py
AngusP:
```python
def _random_good_hashslot_key(self, hashslot):
    """
    Generate a good random key with a low probability of collision between any other key.
    """
    # TODO: Check if the key exists or not. continue to randomize until a empty key is found
    random_id = "{{{0}}}{1}".format(hashslot, self._random_id())
    return random_id
```

You were passing two things to ``format`` but the format string only took one, as `{{0}}` is escaped. The version I've committed will produce:
```
'{hashslot}self._random_id'
```
Whereas the previous one *always* hashtagged into tag `0`:
```
'{0}self._random_id'
```
Ignoring the `hashslot` argument

rediscluster/utils.py
AngusP: I could move them to `client.py`, which is where they are in redis-py. 

I'm avoiding placing this logic in the methods themselves as that removes any ability to use `set_response_callback` to change the command's behaviour. In most cases as well, `execute_command` should return a parsed/correct response as it does with all other commands parsed by callback.

tests/test_pubsub.py
AngusP: It's the same name as the test case in Redis Py, so for symmetry I kept it -- The name is a bit of a mouthful and made in jest, though it does Test the Pub/Sub system's PUBSUB subcommands, so it is descriptive

tests/test_pubsub.py
AngusP: Will do

tests/test_pubsub.py
AngusP: Ok, removed

rediscluster/client.py
AngusP: Also perhaps there's a conflation of hashslot and hashtag terminology here? Hashtags are the part within `{ }` according to the cluster spec. THis sin't relevant to the PR so I can just remove it.

rediscluster/utils.py
AngusP: Also `parse_cluster_nodes` is defined in `utils.py`, which of course is only used by `CLUSTER NODES` and `CLUSTER SLAVES`.

I put them in there mostly as I assumes it was the place you'd want them, given that's where all the other parsers are. Same with redis-py

```python
from .utils import (
    bool_ok,
    string_keys_to_dict,
    dict_merge,
    blocked_command,
    merge_result,
    first_key,
    clusterdown_wrapper,
    parse_cluster_slots,
    parse_cluster_nodes,
    parse_pubsub_channels,
    parse_pubsub_numsub,
    parse_pubsub_numpat,
)
```
It also keeps this nice and neat at the top of `client.py`

tests/test_pubsub.py
AngusP: Done

AngusP:
@Grokzen I've pushed changes based on review, my comments on the review reflect what has changed.

Grokzen:
@AngusP Merging this now. Even with the failing CI tests becuase it looks like 4.0 is also broken when building redis from source. 3.0 works so i am basing that this functionality works in that scenario. If failing tests pop up in the future in 3.2 and 4.0 we deal with them at that time.

Grokzen:
@AngusP Noticed after i merged that you pointed this MR to master and not unstable branch -_- I will cherry-pick all changes into unstable branch manually as that is where they belong

PR #186, opened 2017-02-08
Add support keys that are already unicode encoded and for keys of other types

VascoVisser:
We bumped into a bug where an already encoded unicode string with unicode characters could not be processed by the keyslot function.

While fixing the bug we also added support for other key types such as float, int, long.

The PR include additions to the unit test for the keyslot function

There was another pull request (#170) that is overlapping, but that didn't cover all cases, had no unit test and was pending since beginning of December. This in PR includes the fix in #170.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -22,3 +22,5 @@ Authors who contributed code or testing:
  - monklof - https://github.com/monklof
  - dutradda - https://github.com/dutradda
  - AngusP - https://github.com/AngusP
+ - VascoVisser - https://github.com/VascoVisser
+
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -4,6 +4,7 @@ Release Notes
 Unstable
 --------
 
+    * Fixed issues with some key types in `NodeManager.keyslot()`.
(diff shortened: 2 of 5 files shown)


Grokzen:
@VascoVisser Looks good from my pov. The 2 things left to do is to add a line in release notes and (optional) add your self to authors list. Will be easily merged after those changes.

VascoVisser:
When is this PR planned for merging and release?

Grokzen:
@VascoVisser I tried to resolve the conflict that was still present from the webb gui and it caused more issues and a wierd merge commit. I will manually merge this MR later tonight and possibly make a new release becuase there is nothing more pending that i wait for to be able to make a release.

Grokzen:
@VascoVisser Manually merged these commits and resolved the minor conflict that was still present from a seperate PR that was merged before.

Commits:
- https://github.com/Grokzen/redis-py-cluster/commit/1b175368a96e41d9826b886fcb084e947f1d23d8
- https://github.com/Grokzen/redis-py-cluster/commit/06a9c15975bcbf7b760130ac9966ee3439ab8417

Travis testing of 4.0 still compalins becuase the building there is still kinda broken.

Grokzen:
Many thanks for the patch :)

PR #197, opened 2017-05-18
fix syntax error in pipline-incrby

lostsquirrel:
fix a syntax error might be caused by copy paste


Grokzen:
@lostsquirrel nice find and thands for submiting the patch :)

PR #201, opened 2017-06-21
Add watch, unwatch, mul method to pipeline

max8899:
Add watch, unwatch, mul method to pipeline

    Use transaction with pipeline in redis cluster

    Do the operation with keys with same prefix, so that command will be
    executed on the same node

    * Reference pipeline function

    Signed-off-by: Lei Gong <xue177125184@gmail.com>

### DIFF
--- a/rediscluster/pipeline.py
+++ b/rediscluster/pipeline.py
@@ -201,6 +201,9 @@ class StrictClusterPipeline(StrictRedisCluster):
             if n.connection in self.connection_pool._available_connections.get(node_name, []):
                 self.connection_pool._available_connections[node_name].remove(n.connection)
 
+            if self.connection_pool._created_connections_per_node.get(node_name, 0):
+                self.connection_pool._created_connections_per_node[node_name] -= 1
+
         # if the response isn't an exception it is a valid response from the node
         # we're all done with that command, YAY!
         # if we have more commands to attempt, we've run into problems.
--- a/rediscluster/pipeline.py
+++ b/rediscluster/pipeline.py
@@ -14,6 +14,7 @@ from .utils import clusterdown_wrapper, dict_merge
 from redis import StrictRedis
 from redis.exceptions import ConnectionError, RedisError, TimeoutError
 from redis._compat import imap, unicode
+from .utils import parse_del
 
 
 ERRORS_ALLOW_RETRY = (ConnectionError, TimeoutError, MovedError, AskError, TryAgainError)
@@ -34,7 +35,7 @@ class StrictClusterPipeline(StrictRedisCluster):
         self.startup_nodes = startup_nodes if startup_nodes else []
         self.nodes_flags = self.__class__.NODES_FLAGS.copy()
         self.response_callbacks = dict_merge(response_callbacks or self.__class__.RESPONSE_CALLBACKS.copy(),
-                                             self.CLUSTER_COMMANDS_RESPONSE_CALLBACKS)
+                                             self.CLUSTER_COMMANDS_RESPONSE_CALLBACKS, {'DEL': parse_del})
 
     def __repr__(self):
         """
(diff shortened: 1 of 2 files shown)


max8899:
@Grokzen When could my PR be reviewed? It's kind of block my features which be reviewed in other projects.

Grokzen:
@max8899 Adding these methods and specially to the pipeline support requires detailed review before i will merge this. Adding these methods might not be as easy as it looks.

One thing that i would require before mering is tests of this functionality. There should be tests inside `redis-py` library that you can copy over here and use. The tests from that lib should work on this cluster client. If not then you might have a implementation bug.

Will review this sometime late next week.

rediscluster/pipeline.py
```diff
@@ -162,6 +163,10 @@ def send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=T
 
             # now that we know the name of the node ( it's just a string in the form of host:port )
             # we can build a list of commands for each node.
+
+            if c.args[0] in ['MULTI', 'UNWATCH']:
```
Grokzen: The comments that is above this belongs to the line below `node_name = node['name']`. Move the comments down or move this new if case up above the comments.

rediscluster/pipeline.py
```diff
@@ -34,7 +35,7 @@ def __init__(self, connection_pool, result_callbacks=None,
         self.startup_nodes = startup_nodes if startup_nodes else []
         self.nodes_flags = self.__class__.NODES_FLAGS.copy()
         self.response_callbacks = dict_merge(response_callbacks or self.__class__.RESPONSE_CALLBACKS.copy(),
-                                             self.CLUSTER_COMMANDS_RESPONSE_CALLBACKS)
+                                             self.CLUSTER_COMMANDS_RESPONSE_CALLBACKS, {'DEL': parse_del})
```
Grokzen: Why do you need this? I do not see a reason for it elsewhere in the code.

rediscluster/pipeline.py
```diff
@@ -262,12 +276,18 @@ def load_scripts(self):
...
+
+        return self.execute_command('WATCH', *names)
 
-    def unwatch(self):
+    def unwatch(self, *names):
         """
         """
-        raise RedisClusterException("method unwatch() is not implemented")
+        if not names:
+            raise RedisClusterException("UNWATCH command needs at least on name in this pipeline")
```
Grokzen: Possibly change the error message to `UNWATCH command needs at least one name when used in a pipeline`

rediscluster/pipeline.py
```diff
@@ -262,12 +276,18 @@ def load_scripts(self):
     def watch(self, *names):
         """
         """
-        raise RedisClusterException("method watch() is not implemented")
+        if not names:
+            raise RedisClusterException("WATCH command needs at least on name in this pipeline")
```
Grokzen: Possibly change the error message to `WATCH command needs at least one name when used in a pipeline`

rediscluster/utils.py
```diff
@@ -253,3 +253,13 @@ def parse_pubsub_numsub(command, res, **options):
     for channel, numsub in numsub_d.items():
         ret_numsub.append((channel, numsub))
     return ret_numsub
+
+
+def parse_del(response, *args, **kwargs):
+
```
Grokzen: No blank line here. Remove blank line or add empty docstring

rediscluster/pipeline.py
NoneGG: Should commands on a single node or all commands be unwatched if exception raised?

rediscluster/pipeline.py
```diff
@@ -193,13 +198,19 @@ def send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=T
         # a mismatched result. (not just theoretical, I saw this happen on production x.x).
         for n in nodes.values():
             self.connection_pool.release(n.connection)
+            if n.connection in self.connection_pool._available_connections.get(node_name, []):
+                self.connection_pool._available_connections[node_name].remove(n.connection)
+
+            if self.connection_pool._created_connections_per_node.get(node_name, 0):
+                self.connection_pool._created_connections_per_node[node_name] -= 1
```
NoneGG: I am not sure, but should a connection be discarded directly? The connection pool then seems to be meaningless

Grokzen:
@max8899 Bump on this PR, i still have review comments above that is not dealt with. I will probably not merge this into the next release but the one after that if comments is looked on and fixed.

adolphlwq:
Need `watch` func too. Hope these functions be released as soon as possible.

Grokzen:
Closing this issue as it is not active any longer and the work is to far behind to be implemented right now. A new set of work for this feature would have to be done.

PR #202, opened 2017-06-22
Update client.py

hzhzh007:
if set refresh_table_asap after initialised then  it will case many connection_pool.nodes.initialize() as the client is shared between multiple threads (the shared param refresh_table_asap is seted to false after nodes.initialize()) ，cause machine memory exhausted. And when a slot MovedError, it not need  to initialize all the node.


hzhzh007:
May case some other problem. let it un merged.

Grokzen:
@hzhzh007 I will not merge this as you mentioned.

PR #210, opened 2017-08-04
change self.get to getattr in rediscluster/client.py _get_single_item

Seojiwoong:
change self.get to getattr in rediscluster/client.py _get_single_item


Grokzen:
@Seojiwoong I do not get at all why this should be changed as the current code works as expected. For me to merge this you would have to elaborate on why this is needed and possible provide a failing case when this breaks.

Grokzen:
@Seojiwoong Bump for update on this issue. I will close this one soon if no response is given to my questions, Thanks.

PR #220, opened 2017-09-21
Fix retries for connection errors in pipelines

AlexBoyd:
Fix retries for connection errors in pipelines

When a connection or timeout error is received
when reading out the responses from a redis node
during a pipelined command, all of the commands
from that node need to be marked for retry.

The code was already written to handle this case,
but due to a small oversight in scoping there is
an early return on the first iteration of the loop
that causes all but the first command in the pipeline
for each node not to be marked for retry.

### DIFF
--- a/rediscluster/pipeline.py
+++ b/rediscluster/pipeline.py
@@ -430,6 +430,6 @@ class NodeCommands(object):
                 except (ConnectionError, TimeoutError) as e:
                     for c in self.commands:
                         c.result = e
-                        return
+                    return
                 except RedisError:
                     c.result = sys.exc_info()[1]


AlexBoyd:
Here's a quick python script to repro the issue. It will assert before the patch and not afterward. 

"""
import time

from rediscluster import StrictRedisCluster

# Assumes a local redis-cluster booted up via docker-redis-cluster
startup_nodes = [{"host": "localhost", "port": "7000"},
                 {"host": "localhost", "port": "7001"},
                 {"host": "localhost", "port": "7002"},
                 {"host": "localhost", "port": "7003"},
                 {"host": "localhost", "port": "7004"},
                 {"host": "localhost", "port": "7005"}]

rc = StrictRedisCluster(startup_nodes=startup_nodes)

entryKeys = ['key1', 'key2', 'key3', 'key5', 'key6']
pipe = rc.pipeline()

rc.config_set(name='timeout', value=10)

# By sleeping for more than the timeout configured on the cluster
# the following pipelined commands will all initially receive a timeout error
# and be forced to hit the pipeline'd retry code path.
time.sleep(15)

for entryKey in entryKeys:
    pipe.hmget(entryKey,
               'field1',
               'field2',
               'field3',
               'field4')

results = pipe.execute()
print results

for result in results:
    # hmget with 4 fields should also return an array of 4 values
    assert result == [None, None, None, None]
"""

AlexBoyd:
Just realized this is already a dupe of https://github.com/Grokzen/redis-py-cluster/pull/212. Hopefully this repro script will still be of use though.

Grokzen:
@alexboyd I merged the other Mr #212 as this was a dupe of that one.

Closing this one as dupe of #212.

PR #237, opened 2018-01-10
fix: maximum recursion depth exceeded while calling pfcount

myersguo:
recursion depth exceeded while calling pfcount


myersguo:
@Grokzen , check this issue?

Grokzen:
@myersguo I do not really understand why this change would make any difference? Is the __class__ variable any different from the new definition you have in this PR?

Grokzen:
@myersguo bump on this issue

PR #239, opened 2018-01-12
Python 3.6

PCManticore:
From Travis it seems that redis-py-cluster already support Python 3.6. I just added a new environment in `tox`, as well as a new Trove classifier. Also I had to use a list for the classifiers, since they don't appear on PyPi if they are not a list.


rjauquet:
@Grokzen any update on this PR?

Grokzen:
@PCManticore Thanks for this contribution

PR #218, opened 2017-09-13
Fix #217

AngusP:
Adds `cluster-bus-port` to each node's dict for both >=4.0 nodes and <4.0 to keep consistency

PR #242, opened 2018-01-29
Support 4.0 cluster_nodes with cluster port

YongMan:
redis 4.0 cluster nodes output with cluster bus port.

Fix int(port) fail with port(at)cport and add cport in node if avaliable.


Grokzen:
@YongMan I merged the fix in #218 that solves the same problem. Closing this one as duplicate.

PR #243, opened 2018-02-28
Support passing a specific connection class to StrictRedisCluster 

ilr39:
This cannot be done by using kwargs because kwargs are later passed on to the super init(which is StrictRedis) and it doesn't support arbitrary kwargs, so it fails when trying to pass connection_class as named argument to StrictRedis constructor.

StrictRedis constructor:
` 
def __init__(self, host='localhost', port=6379,
                 db=0, password=None, socket_timeout=None,
                 socket_connect_timeout=None,
                 socket_keepalive=None, socket_keepalive_options=None,
                 connection_pool=None, unix_socket_path=None,
                 encoding='utf-8', encoding_errors='strict',
                 charset=None, errors=None,
                 decode_responses=False, retry_on_timeout=False,
                 ssl=False, ssl_keyfile=None, ssl_certfile=None,
                 ssl_cert_reqs=None, ssl_ca_certs=None,
                 max_connections=None):
`

This doesn't break any interface.
I need this to use SSL connection class with my cluster connection.

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -126,7 +126,8 @@ class StrictRedisCluster(StrictRedis):
     }
 
     def __init__(self, host=None, port=None, startup_nodes=None, max_connections=32, max_connections_per_node=False, init_slot_cache=True,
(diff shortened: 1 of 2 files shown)


Grokzen:
@ilr39 Great find and thanks for the contribution

PR #248, opened 2018-03-23
Remove the magic value 'max_connections=32'

leeglezhao:
Remove the magic value 'max_connections=32' in client.py,using None as default.
If we do not set 'max_connections', it is easy to throw out the exception "Too many connections".


Grokzen:
@leeglezhao Please do not open and close the same PR

Grokzen:
@leeglezhao I am not sure i get why you want this merged, could you elaborate some more why this should be changed to None. A use-case, a reason, an example or similar.

PR #258, opened 2018-06-02
WIP: Respect `readonly_mode` parameter when creating connections via `from_url` method

astawiarski:


astawiarski:
This is odd, these tests doesn't seem to fail because of my changes. Should I base off `master` branch instead of `unstable`?

Grokzen:
@astawiarski No the test errors is from some other fixes i did that did not work out so well on travis-ci but better locally -_-

I looked it over and it is a good MR. It has everything i want from a MR. Small enough change to not brek anything, fixes a problem, comments and migration information. Well done :]

astawiarski:
Nice, thanks for merging. And even more thanks for this great library!

PR #259, opened 2018-06-09
Refresh node table when executing commands on specific nodes

mc3ander:
If the cluster state changes before executing a command that operates on a specific set of nodes (like ping or eval) the client will just retry on the same set of nodes until the clusterdown_wrapper reraises the error. This PR just adds the behaviour from `execute_command` to `_execute_command_on_nodes`. It also changes NodeManager to handle ClusterDown and MasterDown the same way it already handles connection errors, so that it can still initialize a cluster in the case of downed nodes.

### DIFF
--- a/rediscluster/exceptions.py
+++ b/rediscluster/exceptions.py
@@ -74,3 +74,8 @@ class MovedError(AskError):
     """
     """
     pass
+
+class MasterDownError(ClusterDownError):
+    """
+    """
+    pass
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -10,7 +10,7 @@ from .exceptions import RedisClusterException
 # 3rd party imports
 from redis import StrictRedis
 from redis._compat import b, unicode, bytes, long, basestring
-from redis import ConnectionError
+from redis import ConnectionError, ResponseError
 
 
 class NodeManager(object):
@@ -154,10 +154,6 @@ class NodeManager(object):
     def initialize(self):
         """
         Init the slots cache by asking all startup nodes what the current cluster configuration is
-
-        TODO: Currently the last node will have the last say about how the configuration is setup.
-        Maybe it should stop to try after it have correctly covered all slots or when one node is reached
-        and it could execute CLUSTER SLOTS command.
         """
         nodes_cache = {}
         tmp_slots = {}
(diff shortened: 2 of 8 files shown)


tests/test_cluster_obj.py
```diff
@@ -450,3 +450,49 @@ def test_access_correct_slave_with_readonly_mode_client(sr):
...
+                            'port': 7007,
+                            'name': '127.0.0.1:7007',
+                        }]
+                init_mock.side_effect = map_7007
+
+            parse_response_mock.side_effect = side_effect
+            init_mock.side_effect = side_effect_rebuild_slots_cache
+
+            rc = StrictRedisCluster(host='127.0.0.1', port=7006)
+            rc.ping()
```
Grokzen: This test is probably not enough right now. Yes you are testing that internals is used and called, but i do not see any form of validation that it behaves how you expect it to behave. No check if the method is called, no validation that the nodes is rebuilt according to test data etc

rediscluster/nodemanager.py
```diff
@@ -154,10 +154,6 @@ def get_redis_link(self, host, port, decode_responses=False):
     def initialize(self):
         """
         Init the slots cache by asking all startup nodes what the current cluster configuration is
-
```
Grokzen: Why remove this comment? It did not feel like you solved it with your new exception catch?

rediscluster/exceptions.py
```diff
@@ -74,3 +74,8 @@ class MovedError(AskError):
     """
     """
     pass
+
+class MasterDownError(ClusterDownError):
```
Grokzen: New exception class but unused right now. It is imported in previous file but not used there.

Grokzen:
Looks good otherwise, i just want to think some things through wile you fix the comments and before i merge this in.

rediscluster/exceptions.py
mc3ander: I've added it to the exception parser in connection.py. It's a subclass of ClusterDownError, so it'll just be treated the same throughout the client, since the cause of both errors are pretty much the same.

rediscluster/nodemanager.py
mc3ander: I didn't change the behaviour, but it looks like it's already been solved with the break on line 239.

Grokzen: Right, i see what you are refering to, great catching this

Grokzen:
@mc3ander Do you have additional changes you want to push up in this PR or can i start to consider to merge this? I think it is mergable as is right now, just want to check with you.

mc3ander:
I've got other changes, but they still need some clean up and tests, so they're probably best as separate PRs. This should be good to go. Thanks

Grokzen:
@mc3ander the github merge conflict resolver did not work out and it would not let me fast forward it here so i had to manually merge the changes into unstable branch. Thanks for this contribution.

Grokzen:
8bfe9c8..bef416b is the merged commits in unstable branch.

PR #261, opened 2018-06-14
Issue 260 unexpected keyword argument ssl

arthur78:
Please have a look. This change works for me on AWS.


Grokzen:
@arthur78 I think i beat you to it :] I fixed it in a very similar way in this commit b12c3538b3b1c46e7ba7ce9f78c245a80ced5f7f and it basically does the same thing your code do but the `else` clause is already present in the connection pool class. Thanks for reporting this

PR #272, opened 2018-08-22
Fix Error: TTL exhausted.

copyangle:
sometimes
if slot_id == wrong : 
    while ttl == nothing and MOVE == nothing

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -392,6 +392,7 @@ class StrictRedisCluster(StrictRedis):
 
                 node = self.connection_pool.nodes.set_node(e.host, e.port, server_type='master')
                 self.connection_pool.nodes.slots[e.slot_id][0] = node
+                slot = e.slot_id if e.slot_id != slot else slot
             except TryAgainError as e:
                 if ttl < self.RedisClusterRequestTTL / 2:
                     time.sleep(0.05)


Grokzen:
@copyangle You need to provide much more detailed information on exactly when this happen, the entire scenario, cluster setup, example script to trigger the error, procedure when this happens if there is any failover or similar that happens.

copyangle:
For example like this.
pieces = ["get","save"]
REDIS_CLUSTER.execute_command('CONFIG',*pieces)

Grokzen:
Are you creating your rediscluster instance with any different settings other than usual? Is only `CONFIG` causing these problems or are other commands doing it as well?

One problem i am thinking of is that you are trying to run `CONFIG` that is not really a slot based command and should be routed to all masters if i remember correct. `CONFIG` should not have to route to a specific slot.

One possible solution is that you should run something like `.execute_command('CONFIG GET', *pieces)` and that might help you to route the command properly becuase my guess is that the command you are trying to send out now is not present in the override config here https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/client.py#L55 and if your command is not detected in that config, then it will assume the command is a slot based command and try to route you to a node based on your keys by doing a slot lookup. Please try to change this and see if that improves anything for you.

copyangle:
Yes, I am trying to extend it. But I think it should have such scalability, not an error.

Grokzen:
@copyangle What is the acctual usecase for this feature? Unless you can provide a good reason for it, i see no reason to merge this as you are using it wrong based on how both redis-py have implemented it and how i have implemented it here.

copyangle:
When the MOVE command appears and the following code is executed:
` 
           except MovedError as e:
                # Reinitialize on ever x number of MovedError.
                # This counter will increase faster when the same client object
                # is shared between multiple threads. To reduce the frequency you
                # can set the variable 'reinitialize_steps' in the constructor.
                self.refresh_table_asap = True
                self.connection_pool.nodes.increment_reinitialize_counter()

                node = self.connection_pool.nodes.set_node(e.host, e.port, server_type='master')
                self.connection_pool.nodes.slots[e.slot_id][0] = node

`
A new node is recorded and assigned. But the second round of execution will override this assignment, which is equivalent to not changing anything. If I understand that there is an error, please correct me, thank you.

` 
               if self.refresh_table_asap:
                    # MOVED
                    node = self.connection_pool.get_master_node_by_slot(slot)
                else:
                    node = self.connection_pool.get_node_by_slot(slot)
                r = self.connection_pool.get_connection_by_node(node)

`

Grokzen:
So you mean basically that you need to have the old slot assignment before the slot:node table was reinitialized becuase the exception tries to update the slot to node assignment, but you want it to use the one it had mapped before the exception?

copyangle:
No, I want him to use the new node returned by the system after an exception occurs. In this case, the slots for the redis cluster calculation and the slots for the redis cluster calculation are not consistent. By modifying the slot value, MOVE takes effect on cluster.

Um...... Perhaps my understanding is not enough, I just use this method to complete the function I need. : )

Grokzen:
But as i said before, there should be no MOVED error possible when you are running CONFIG commands because according to this line https://github.com/Grokzen/redis-py-cluster/blob/unstable/rediscluster/client.py#L55 `CONFIG GET` & `CONFIG SET` is commands that should be sent to all masters in the cluster and returned as a dict for each node and/or master in the cluster. The `CONFIG` command should not do any node or slot lookup because all it should lookup is what master nodes to send to and sending a command to a master node should not throw a `MovedError` as moved errors is only applicable to slot based commands.

If you add `CONFIG` to the list of items on L55 as i lined before and try again, your command that you run should properly route your command to all master nodes and you should get back the data that you request properly.

On a side note, i do not see why you simply can't run?
```
pieces = ["save"]
REDIS_CLUSTER.execute_command('CONFIG GET',*pieces)
```

Becuase this is the way to use the command. I might accept a change to add `CONFIG` to the `NODES_FLAGS` & `RESULT_CALLBACKS` to make it work. But i would need a better reason why this should be done in this way over the code i posted above that is the proper/correct way of doing it.

Grokzen:
@copyangle bump on this issue

PR #292, opened 2018-11-12
bug in initialization of redis cluster slots, fixed

afshinrudgar:
Hi
I happened to stumble upon this error with `initialize` phase of RedisCluster, with fetching cluster nodes the starting node with send `127.0.0.1` as host, and it should be changed to node host address.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ build
 .cache
 docs/_build
 docs/_build_html
+.idea
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -3,14 +3,14 @@
 # python std lib
 import random
 
+# 3rd party imports
+from redis import ConnectionError, TimeoutError, ResponseError
+from redis import StrictRedis
+from redis._compat import b, unicode, bytes, long, basestring
+
 # rediscluster imports
 from .crc import crc16
 from .exceptions import RedisClusterException
-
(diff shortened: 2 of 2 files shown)


Grokzen:
@afshinrudgar I see a lot of lint fixes and some feature fix inside a single commit. Can you split them up for easier review?

Grokzen:
Also i do not really get what you are trying to solve here? I would argue that it is not up to this lib to magically change the host if the redis-cluster itself do not report back the correct ip/name back to the clients. I do not see me merging this becuase of that reason.

If your redis-cluster reports back 127.0.0.1 for your nodes then you probably have used 127.0.0.1 as the ip address that you connect to when your nodes try to connect to the master. If you instead use an externally available IP address to point to your cluster nodes, redis-server will record this and clients will get back that information from `CLUSTER INFO` command and connect to your cluster correct. I had this issue when running redis-server inside a single docker container, see my other repo docker-redis-cluster.

afshinrudgar:
Yes. you're right.
I used 127.0.0.1 in my setup, I just considered that `redis` would handle 127.0.0.1 and changed that to host IP or it was just some sort of bug on their side.
thanks

Grokzen:
I know, it was more difficult in the past to get the server to set the correct ip you wanted it to present to your clients. But it is not up to the client to solve this issue.

PR #293, opened 2018-11-12
Bugfix/redis4 cluster initialization

afshinrudgar:
Bug in redis cluster initialization with redis 4


Grokzen:
@afshinrudgar  Why open a second Pull request when you had one open? Also you have some problems with where you  created your brnch from as you should not see all of these `merged branch unstable` commits in your pull request-

Grokzen:
@afshinrudgar Again, please read the last 2 comments in your previous pull request before you continue with this pull request and anser it. I am currently not inclined to merge this PR as it solves a problem in the wrong way/location. Please respond to either PR with why this should be done here and not solved with a proper depoyment of the redis-server. Without a good reason for that question, i probably won't merge this.

Issue #295, opened 2018-11-16
Update codebase to work with redis-py 3.0.0


Grokzen:
The latest release of redis-py 3.0.0 have several major breaking changes that needs to be ported and verified in this repo. For now the 1.3.x track will only support 2.10.6 and the plans is that when 1.4.0 is out it will only work with redis-py 3.0.0. If redis-py continues to support the 2.x version line the 1.3.x track in this repo will keep it supported.

caiohenriquesilva:
The problem is the requirements file that is installing the equal or major version than 2.10.6 imo. In my project requirements I had to force the 2.10.6 installation for redis to redis-py-cluster works.

Grokzen:
@caiohenriquesilva I do not get what you mean? The latest release 1.3.6 solved this issue by hardcoding the dependency to `==2.10.6` to solve the root cause of it. That has nothing really todo with this issue right now as the immediate problem was solved.

CheViana:
would love this released asap & ready to help

BRIDGE-AI:
is this on going?
I'd like to get the status of this, plz.

Grokzen:
Release 2.0.0 is done and published

PR #296, opened 2018-11-26
WIP: redis-py 3.0.x support

Grokzen:
This branch is the current WIP to merge and update all code to make it compatible with redis-py >= 3.0.0,<3.1.0 releases. Current plan is to sort out all the broken stuff and make it work again with most basic commands before moving to support all new commands and features as that needs more exploration on how they work in a clustered environment.

Solves #295


CheViana:
would love to have these changes, ready to help

Grokzen:
@CheViana It is moving along good. The features that i have no clue about right now is the latest new features and commands that was added into redis 5.0.x series. Streams for example. I have no idea how they perform or work or behave in a clustered envrionment and if they are really ment to be supported by a cluster or not. I got most other things covered right now and it do not seems to be that difficult to deal with, only the backwards incompatible things is the real issue in a way.

yakhira:
it seems to be stuck year ago

Grokzen:
Most tests now work as expected or have had a skip block added to them either because of crossslot reasons or they should hardfail. Noteable change is that i ripped out the sort method and will make a fallback onto the old crossslot error provided by redis itself.

PR #305, opened 2019-01-10
Fix "Target key name is busy" error when key already exists

vsouza:
According Redis documentation __RESTORE__ command will return a "Target key name is busy" error when key already exists unless you use the __REPLACE__ modifier. I'm using this project in production and facing errors, because my pipeline performs an __RENAMENX__ operation.

The `redis-py` driver implements the replace option in :

```python
    def restore(self, name, ttl, value, replace=False):
        """
        Create a key using the provided serialized value, previously obtained
        using DUMP.
        """
        params = [name, ttl, value]
        if replace:
            params.append('REPLACE')
        return self.execute_command('RESTORE', *params)
```
[link](https://github.com/andymccurdy/redis-py/blob/1244f6e3f9d38379124622ff4f0371914ce759ae/redis/client.py#L1400-L1408)

So I made a change in `client.py` enabling a replace option.

Thank you.

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -726,7 +726,7 @@ class StrictRedisCluster(StrictRedis):
 
         return self.mset(**kwargs)
 
-    def rename(self, src, dst):
+    def rename(self, src, dst, replace=False):
         """
         Rename key ``src`` to ``dst``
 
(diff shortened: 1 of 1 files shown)


Grokzen:
@vsouza I will probably drag this into the redis 3.0.0 branch i have open to include them both in the same release later on. Good find

vsouza:
Thanks @Grokzen I'll use my fork until the release of 3.0.0 branch.

Grokzen:
@vsouza I manually merged this into the redis-3.0 branch. Thank you for this contribution.

PR #306, opened 2019-02-06
Clearer exception message when cluster slots fails

imnotjames:
it was difficult to get to the bottom of the cause of the failed
cluster slots command messages so this adds the cause message into
the payload and adds the cause as a property

### DIFF
--- a/rediscluster/exceptions.py
+++ b/rediscluster/exceptions.py
@@ -75,7 +75,20 @@ class MovedError(AskError):
     """
     pass
 
+
 class MasterDownError(ClusterDownError):
     """
     """
     pass
+
+
+class ClusterSlotsCommandError(RedisClusterException):
+    """
+    """
+    def __init__(self, node, cause):
+        super(ClusterSlotsCommandError, self).__init__(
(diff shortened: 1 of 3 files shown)


Grokzen:
@imnotjames Why is it good to encapsulate the `node` object into this exception when raising it? I see no real benefit from propegating it further up the stack to the user of the lib.

imnotjames:
I suppose we could remove it? It would help if you were to programmatically utilize it as part of a custom node manager, connection pool, or otherwise.

Grokzen:
I have yet to see anyone post or mention or show a custom node manager yet for a cluster scenario. Not saying there is none, just that i have not seen one yet.
I guess it do not really hurt at all, but it is more of the part that it might not be a piece that ever will have a good use or a good use-case for what scenario it actually might help out in some way. Making a more specific exception class might be good to make it more obvious what the error would be.

imnotjames:
More specific exception classes would be great across all of the cluster exceptions :)

We're using this in a nodemanager to ensure we keep unhealthy nodes out of the pool.

PR #307, opened 2019-02-06
Handle ResponseError better in Python 3

### LINKED ISSUE #278: 'ResponseError' object has no attribute 'message'
nodemanager.py - > line 180. ResponseError object tries to use message attribute, but message attribute was mark deprecated in 2.6 and was removed in further versions

imnotjames:
Previously, we try to hit `ResponseError.message` to check if this is a `MASTERDOWN` or otherwise.

However, in Python 3, there's no `message` and instead we just need to cast the error to a string.

This should fix #278

### DIFF
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -177,7 +177,8 @@ class NodeManager(object):
                 continue
             except ResponseError as e:
                 # Isn't a cluster connection, so it won't parse these exceptions automatically
-                if 'CLUSTERDOWN' in e.message or 'MASTERDOWN' in e.message:
+                message = e.__str__()
+                if 'CLUSTERDOWN' in message or 'MASTERDOWN' in message:
                     continue
                 else:
                     raise RedisClusterException("ERROR sending 'cluster slots' command to redis server: {0}".format(node))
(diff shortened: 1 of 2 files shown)


AvnerCohen:
Is this planned to be merged?  Anything blocking?
Would love to see the project better adapting python3 as we get closer to 2's EOL.

Grokzen:
@AvnerCohen This project will follow the same python support matrix that the upstream project `redis-py` uses, i will not deviate from that to support more or less python releases. Dropping 2.6 a while back as they did that in a release for example. If you want to further refine/update the python support then make it happen in upstream first and that will propagate down to this project eventually.

Grokzen:
@AvnerCohen This project will follow the same python support matrix that the upstream project `redis-py` uses, i will not deviate from that to support more or less python releases. Dropping 2.6 a while back as they did that in a release for example. If you want to further refine/update the python support then make it happen in upstream first and that will propagate down to this project eventually.

AvnerCohen:
@Grokzen Maybe something I am missing here.
`redis-py` formally supports `Python 3.4+`.
This pull request is meant to *allow*  python 3.x support.

So it's fully inline and in sync with up stream, isn't it ?

PR #308, opened 2019-02-07
Use Binary and Unicode literals instead of the helpers from redis-py

imnotjames:
These helpers are being removed in the next version of `redis-py` (`>=3.0`).  With that in mind, I cleaned up and converted everything to the same systems that are used by the new version of `redis-py`.

This should help us upgrade to `redis-py >= 3.0`

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -36,7 +36,7 @@ from .utils import (
 from redis import StrictRedis
 from redis.client import list_or_args, parse_info
 from redis.connection import Token
-from redis._compat import iteritems, basestring, b, izip, nativestr, long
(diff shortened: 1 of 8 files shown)


imnotjames:
That's weird - it failed the checks because of some AWS issues?

Grokzen:
What AWS issue? This repo do not use any AWS features? Unless you mean on the travis side itself or that coveralls is broken?

imnotjames:
Yeah, it was on the travis side of things - it couldn't retrieve the blob for travis to run Python 3.4 - but it seems to be cleared up now!

imnotjames:
Rebased against unstable - not sure what went wonky there.

PR #309, opened 2019-02-11
Basic support of RedisPy 3.0

imnotjames:
This implements a very basic support of `redis-py` 3.0.  It relies on #308 being merged.  Once that is merged into unstable I'll rebase it and it'll be easier to read.

Per your branch, I updated it to support `>=3.0` and `< 3.10`.

No new features are supported through this.


docs/release-notes.rst
```diff
@@ -13,7 +13,7 @@ Release Notes
     * Add Redis 4 compatability fix to CLUSTER NODES command (See issue #217)
     * Fixed bug with command "CLUSTER GETKEYSINSLOT" that was throwing exceptions
     * Added new methods cluster_get_keys_in_slot() to client
-    * Fixed bug with `StrictRedisCluster.from_url` that was ignoring the `readonly_mode` parameter
+    * Fixed bug with `RedisCluster.from_url` that was ignoring the `readonly_mode` parameter
```
imnotjames: Should I leave this one be?  I was going through docs and it made sense to change it, 'cause it no longer references the right thing.

imnotjames:
Should this update the commands to match the `RedisCluster` ones?  Looks like `mset` only accepts a `mapping` dictionary now.

Grokzen:
@imnotjames There already is a branch for redis-py 3.0 support, almost all changes here is already implemented in that branch in #296 There is a few things in here i can pull out and use that i have not yet got to, but if i merged this into my branch, most changes would be null.
What i need the most right now is either to do real world tests of my branch, port or update all tests that was pulled from redis-py way back when to ensure they are up-to-date with the latest code, or investigate how the new commands in 5.0 works out in a cluster environment as i have no clue or yet tinkered with them in a cluster env and how they perform and if they are even fesable to run in a cluster env.

imnotjames:
Sure, it's duplicated effort.  Just trying to get something with redis-py 3.0 where tests pass to try it out against beta deployments at my work.  Folks are wanting to use some of the new features available in 3.0. 

I'm currently investigating what tests can be synced over from redis-py 3.0 into this branch. 
 I've been marking the new 5.0 features as "unsupported" and skipped.

imnotjames:
`eae7209` seems to be a general bug in `redis-py-cluster` where the `OBJECT` command is mis-routed.  I'm going to cherrypick it as its own PR.

However, I've updated the tests to match `redis-py` 3.0 as closely as possible and updated them where it made sense given some of the differences between `redis-py` and `redis-py-cluster`

I also marked the stream commands as blocked - the tests didn't seem to be working as expected.  Further work can be done beyond this PR to implement those features.

PR #311, opened 2019-02-25
ensure OBJECT command is directed to correct node

imnotjames:
The `OBJECT` command ends up getting sent to the wrong node.

Because this is an uncommon edge case, I added a special case within `_determine_node` so we correctly find the `key` we're working with when using an `OBJECT` command.

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -285,7 +285,10 @@ class StrictRedisCluster(StrictRedis):
                 raise RedisClusterException("{0} - all keys must map to the same key slot".format(command))
             return slots.pop()
 
-        key = args[1]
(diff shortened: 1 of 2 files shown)


rediscluster/client.py
```diff
@@ -285,7 +285,10 @@ def _determine_slot(self, *args):
                 raise RedisClusterException("{0} - all keys must map to the same key slot".format(command))
             return slots.pop()
 
-        key = args[1]
+        if command == 'OBJECT':
```
Grokzen: A very small thing, but i would prefer it to be this

```
key = args[1]

if command = 'OBJECT':
    key = args[2]
```

Alter this and i merge this PR after that.

PR #315, opened 2019-03-28
Fix typo and syntax error

imnotjames:
There's a syntax error and typo in the client.  This corrects that issue.


imnotjames:
Looks like 3.3 is no longer supported by the setuptools version you're using

PR #316, opened 2019-03-30
drop Python 3.3 support and add Python 3.7 support

imnotjames:
Python 3.3 is no longer supported by the version of setuptools getting pulled in.


imnotjames:
Also, tests were failing already - addressed in #315

Grokzen:
I do not complain about dropping old python versions :) thanks @imnotjames

PR #322, opened 2019-05-28
Add xenial dist to support python 3.7+.

engstrom:
Attempt to fix the failing build on python 3.7.

According to the Travis CI docs, python 3.7 and higher require the `xenial` distribution.
https://docs.travis-ci.com/user/languages/python/#python-37-and-higher

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: false
+dist: xenial
 language: python
 python:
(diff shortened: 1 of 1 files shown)


engstrom:
Hi, @Grokzen. Do you know when the next release might be? I'd love for this to get included and I'm available to assist if there's anything else you need for the next release.

PR #324, opened 2019-06-28
Feature/redis py 3.0

Phanapoch:
fix import error in redis-py >= 3.2


Grokzen:
No, this MR will not be accepted as is. I will not open up to support a major version line of redis-py when the code changes that was introduced in those version lines is not merged and sorted out in a cluster envrionment. It is not as simple as just to change the requirements file and push it out.
The next major version of this lib will support up to redis-py 3.3.x and hopefully be compatible all the way back to 3.0.x versions, but we will see how that goes.

PR #336, opened 2019-10-11
:bug: fix TestSSLConnectionURLParsing and TestBlockingConnectionPool

yangjiaronga:


yangjiaronga:
I found that you had a lot of errors in overwriting the test

Grokzen:
@yangjiaronga For each version of the redis-py packages that i support i merge in all of their tests that they have and adapt any tests that needs adapting to a clustered setting/envrionment. Merging all tests and ensure everything still works and is compatible is one of the most major/biggest things that i need to do between all supported releases. It is on purpose that if they update their tests i need to update mine as well, it is not optimal but it is what i can do right now to ensure that the cluster code works as expected and in the way i want since we override and modify upstream behaviour in many places.

I agree and disagree that i should only focus on redis-cluster behaviour, yes that is the primary goal and things that needs to be tested and validated, but since we muck around with commands and baasically changes how everything is routed around a cluster env, i need to test every single command that i want to be supported, there has been errors in the past that has surfaced due to things that has been done here with very simple commands.

Is this change some port of a change that was done inside reids-py in some version? If so can you please point me to the git commit that introduced these changes?

If you are looking into the master branch and test_lock.py for example, i think that was a file i ported in very recently and i have not goten around to sort all the new files out, and yes, ofc they should be using the rediscluster class instead of the upstream redis class. It is just a minor oversight that it is that right now and it will all be sorted before the next major release.

yusukegoto:
Hi, I think this change has been seen since redis-py v3.3.11.

```
# ensure this connection is connected to Redis
connection.connect()  # HERE!!
```

https://github.com/andymccurdy/redis-py/commit/cfa2bc9b7ea860eb4a002eaa7029ecee01e39735

Grokzen:
Since the code from any release including and above 3.1.0 has not yet been fully merged into the code-base, this change is not really needed righ tnow as i would pick this up when i do merge the code and all related changes from the upstream/future versions of the code. I can keep this open for now, but it will be closed as invalid later on.

Grokzen:
Manually added this to the code base in the following commit https://github.com/Grokzen/redis-py-cluster/commit/f273b65d887a4d711a6b5a4fed1176d7080a5b43 as it includes more changes from upstream redis-py lib. Closing this PR and the fixes requested will be part of next release.

PR #344, opened 2019-11-15
Fixed some basic grammar and spelling

alan-yilun-li:
Hi! While on-boarding onto using this project I decided to fix some of the spelling and grammar I came across.

Please let me know changes like this are unwanted and I'll stop with this. Thanks for making this available!


Grokzen:
@alan-yilun-li Documentation is my biggest weakness, so any help with that is always very much appreciated to make it better then before.

PR #345, opened 2019-11-20
try to refresh topology of nodes before totally give up

wnykuang:
In some case the cluster is not dead but one of the nodes come with new IPs.
We should try to refresh the topology and run the query again

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -421,7 +421,7 @@ class RedisCluster(Redis):
                 raise
             except (ConnectionError, TimeoutError):
                 try_random_node = True
-
+                self.refresh_table_asap = True
                 if ttl < self.RedisClusterRequestTTL / 2:
                     time.sleep(0.1)
             except ClusterDownError as e:
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -421,6 +421,7 @@ class RedisCluster(Redis):
                 raise
             except (ConnectionError, TimeoutError):
                 try_random_node = True
+
                 if ttl < self.RedisClusterRequestTTL / 2:
                     time.sleep(0.1)
             except ClusterDownError as e:
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -421,7 +421,6 @@ class RedisCluster(Redis):
                 raise
             except (ConnectionError, TimeoutError):
                 try_random_node = True
-                self.refresh_table_asap = True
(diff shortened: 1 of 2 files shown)


Grokzen:
@wnykuang Can you give a scenario or case where this really happens a lot? For example does this happen a lot when you are using a redis cluster in a cloud envrionment or something similar? Also i think that setting it to trying a random node and refreshing the table asap conflicts with eachother as it would basically reset and make your client lookup the entire cluster again at every time a node drops off and that is usually not something that is wanted by default.

wnykuang:
> @wnykuang Can you give a scenario or case where this really happens a lot? For example does this happen a lot when you are using a redis cluster in a cloud envrionment or something similar? Also i think that setting it to trying a random node and refreshing the table asap conflicts with eachother as it would basically reset and make your client lookup the entire cluster again at every time a node drops off and that is usually not something that is wanted by default.

When we deployed the redis-cluster in kubernetes, sometimes k8s killed one of the redis pods for some reason and will recover in seconds. In that scenario node's IP address be changed and I don't think it will be recovered from try_random_node since the err is on ```https://github.com/Grokzen/redis-py-cluster/blob/master/rediscluster/client.py#L378``` line.

If I have misunderstood any thing, please let me know. Thanks

wnykuang:
> @wnykuang Can you give a scenario or case where this really happens a lot? For example does this happen a lot when you are using a redis cluster in a cloud envrionment or something similar? Also i think that setting it to trying a random node and refreshing the table asap conflicts with eachother as it would basically reset and make your client lookup the entire cluster again at every time a node drops off and that is usually not something that is wanted by default.

Yeah, I agree with the `self.try_random_node` and `self.refresh_nodes_asap` is conflicted to each other.

However, sometimes the IP address of node is changed and the client should refresh the topology. I don't think the decorator will be called many times tough, but I think it should be fine( to refresh the nodes IP when it broken), right?

Grokzen:
@wnykuang I can't really merge this as is. What i want you to do for this to be mergable is to first squash all of your changes into a single commit. Secondly you need to move the code over to client.py now as the method you tried to fix has been moved into client.py into the method `execute_command`. You also need to fix or figure out why tests fail when i tried to add this code it started to fail the tests, more specifically `test_busy_loading_disconnects_socket` that fails for some unknown reason. Until this is done, i won't merge this into the codebase.

PR #347, opened 2019-12-02
Added Clustered Blocking Connection Pool and Accompanying Tests

alan-yilun-li:
Built ClusterBlockingConnectionPool based on the BlockingConnectionPool in redis-py.

Enabled it for clustering following the adjustments made in ClusterConnectionPool.
Adapted the tests for BlockingConnectionPool similarly and added some extra cases.

Fixed some typos along the way.

Currently includes a solution for supporting `max_connections_per_node = False` though it is somewhat naive. I have another PR that only supports `max_connections_per_node = True` which simply removes this option, gating it behind `NotImplementedError` in case that's better.

### DIFF
--- a/rediscluster/connection.py
+++ b/rediscluster/connection.py
@@ -416,14 +416,16 @@ class ClusterBlockingConnectionPool(ClusterConnectionPool):
     def reset(self):
         self.pid = os.getpid()
         self._check_lock = threading.Lock()
-        self._pool_by_node = None
-        self._group_pool = None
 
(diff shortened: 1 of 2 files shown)


Grokzen:
@alan-yilun-li The feature looks good and the tests looks good and well adapted as well. I might just put the blocking pool as a experimental feature in the 2.1.0 release but it still looks solid/stable to use.

And in general, the issue with having multiple pools at the same time is a difficult one to solve indeed.

PR #353, opened 2020-01-29
fix: ClusterConnectionPool.count_all_num_connections is not thread safe in python 3

mattrobenolt:
In python 3, `dict.values()` is an iterator, whereas in python 2 it
returns a copy of the values as a list.

When running under python 3 and multiple threads, that means the
dictionary _created_connections_per_node is now able to be mutated and
iterated over at the same time, thus raising a RuntimeError.

The race comes from multiple threads calling
`ClusterConnectionPool.make_connection` at the same time.

Coercing the `.values()` to a `list()` _appears_ to the best of my
knowledge to be locked by the GIL and doesn't cause this issue.

I have stress tested this with: https://gist.github.com/mattrobenolt/e88d940c9afa6a18d6b39328eeb9a8f6

https://sentry.io/share/issue/7e8780304998400a934c0dbeb9eab888/

### DIFF
--- a/rediscluster/connection.py
+++ b/rediscluster/connection.py
@@ -220,9 +220,10 @@ class ClusterConnectionPool(ConnectionPool):
         """
         Create a new connection
         """
-        if self.count_all_num_connections(node) >= self.max_connections:
+        num_connections = self.count_all_num_connections(node)
+        if num_connections >= self.max_connections:
             if self.max_connections_per_node:
-                raise RedisClusterException("Too many connection ({0}) for node: {1}".format(self.count_all_num_connections(node), node['name']))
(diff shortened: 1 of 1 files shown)


rediscluster/connection.py
```diff
@@ -220,9 +220,10 @@ def make_connection(self, node):
         """
         Create a new connection
         """
-        if self.count_all_num_connections(node) >= self.max_connections:
+        num_connections = self.count_all_num_connections(node)
```
mattrobenolt: This is a subtle change, but just prevents `self.count_all_num_connections` from needing to be called and calculated twice if/when the exception is raise.

Grokzen:
@mattrobenolt Thank you for this contribution

Issue #356, opened 2020-02-24
Reinstate support for python 3.5


n89nanda:
Looks like a commit on Jan 13 removed support for python 3.5. I am trying to understand why this is no longer supported. 

I was able to find below reason for python 3.4. 
`A segfault was found when running `redis-py` in python `3.4.0` that was introduced into the codebase in python itself in the version `3.4.0`.`

Also, the redis-py library still supports python 3.5 and python 3.5 has not reached EOL yet.

Thanks for any additional information on this issue.

Grokzen:
My philosophy for supported python version in the 3.x version series is to support the latest + the 2 versions behind it. I think this is a way that most packages that is built for python3 versions should be supported and done in general. I will give it some thought tho to possible revert that as i have stated before that i want to follow the supported matrix that redis-py uses to preserv compatibility. Thanks for brining it up.

n89nanda:
Thanks for the quick response and thanks again for looking if we can bring back support. I am currently upgrading from 1.3.x to `master` for latest `redis-py` support. It will be great if we can keep it same as `redis-py` since I am using that separately as well.

Grokzen:
Why did you open one, close it, and open the same one anew?

n89nanda:
Oh, Initially I wasn't sure if I made changes on master or one of the previous commits where the python 3.5 support was removed. Also, my email in git-commit was not associated to my github.
Just corrected git config on my end.
Anyway, If you plan to bring back support, please review this PR #358 https://github.com/Grokzen/redis-py-cluster/pull/358

Grokzen:
MR #358 merged, closing issue

PR #358, opened 2020-02-24
added back python 3.5 support

n89nanda:
Related to issue #356 
 https://github.com/Grokzen/redis-py-cluster/issues/356


Grokzen:
@n89nanda After considering it, i will follow what i said before that i will follow the support matrix of the upstream redis-py as much as possible even if it would go outside my normal and usual way of dealing with supported python versions. Thank you for this contribution.

PR #372, opened 2020-05-21
Remove Token class checking when use XREADGROUP/XREAD

bomu-bomu:
redis.connection.Token  was removed from client.py but Token symbol still exists in __determins_slot_ if we use XREAD/XREADGROUP, which cause error when using XREAD/ XREADGROUP 

```
Traceback (most recent call last):
  File "stream-consumer.py", line 16, in <module>
    'stream' : '>'}
  File "/usr/local/lib/python3.6/dist-packages/redis/client.py", line 2607, in xreadgroup
    return self.execute_command('XREADGROUP', *pieces)
  File "/usr/local/lib/python3.6/dist-packages/rediscluster/client.py", line 529, in execute_command
    return self._execute_command(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/rediscluster/client.py", line 561, in _execute_command
    slot = self._determine_slot(*args)
  File "/usr/local/lib/python3.6/dist-packages/rediscluster/client.py", line 467, in _determine_slot
    tokens = {args[i].value: i for i in range(len(args)) if type(args[i]) == Token}
  File "/usr/local/lib/python3.6/dist-packages/rediscluster/client.py", line 467, in <dictcomp>
    tokens = {args[i].value: i for i in range(len(args)) if type(args[i]) == Token}
NameError: name 'Token' is not defined
```

So I replaced these lines with  byte string checking for index number,  instead.

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -464,8 +464,8 @@ class RedisCluster(Redis):
             return slots.pop()
 
         if command in ['XREADGROUP', 'XREAD']:
-            tokens = {args[i].value: i for i in range(len(args)) if type(args[i]) == Token}
-            keys_ids = list(args[tokens['STREAMS'] + 1: ])
+            stream_idx = args.index(b'STREAMS')
+            keys_ids = list(args[stream_idx + 1: ])
             idx_split = len(keys_ids) // 2
             keys = keys_ids[: idx_split]
             slots = {self.connection_pool.nodes.keyslot(key) for key in keys}


Grokzen:
@bomu-bomu I have not been playing around with streams, is it possible that you post an example script or parts of your stream-consumer script that i can run locally to test and use the streams feature? Because what i am seeing is that all `test_x*` methods is currently blocked out because i have not really gone through that feature and tested if it is cluster compatible or not, so i dont know if this change will fix or break anything other then that it solves the most obvious problem that Token is not defined, but i dont know if this breaks or fixes anything else with streams really. Is that something you can provide?

bomu-bomu:
Just publish example scripts  for that: 
https://github.com/bomu-bomu/Redis-stream-on-cluster-sample

Grokzen:
@bomu-bomu Thank you for the script sample, i will dig right into it

Grokzen:
@bomu-bomu Thank you, the fix looks solid, i tested your script and i also re-enabled all the tests that was blocked out and they all pass without any issues or problems.
Thank you for this contribution.

PR #374, opened 2020-06-04
DOCS: Fix spelling and grammar in pipelining documentation

foxyblue:
Maintaining a line limit on these docs might help improve the diff from git, since they hide my most notable correction of `sexist` -> `exist`


Grokzen:
Wow that typo was embarassing to see that it was in there... Thank you for this contribution, documentation fixes is allways appreciated.

PR #378, opened 2020-06-13
[Bug Fix] Ensure port mapping is not in isolation of host mapping

appurvj:
The existing remapping logic looks at host remapping and port remapping in
isolation, which leads to failure in some cases where all from_ports are same
like the following example.

```
startup_nodes = [
	{"host": "127.0.0.1", "port": "17000"},
	{"host": "127.0.0.1", "port": "17001"},
	{"host": "127.0.0.1", "port": "17002"},
	{"host": "127.0.0.1", "port": "17003"},
	{"host": "127.0.0.1", "port": "17004"},
	{"host": "127.0.0.1", "port": "17005"}
]

host_port_remap=[
    {'from_host': '41.1.3.1', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17000},
    {'from_host': '41.1.3.5', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17001},
    {'from_host': '41.1.4.2', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17002},
    {'from_host': '50.0.1.7', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17003},
    {'from_host': '50.0.7.3', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17004},
    {'from_host': '32.0.1.1', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17005}
]
```
This is becasue when just ports get remapped, it ends up producing overlapping node names
and ends up corrupting both the `nodes_cache` and the `slots` dictionaries in nodemanager

Changes:
- This change fixes that issue by nesting the port mapping logic under the host mapping logic.
    The idea is that port mapping should only happen in addtion to host mapping and not without
- Another minor change that is introduced is that the host_port_remap can be thought of as
    rules chain so if we encounteer an entrry to leads to remapping, we dont consider the
    "remapping rules" under that entry in the list.
- Also fixed an unrelated failing testcase `test_blocked_commands` in `test_cluster_obj.py`
- Updated the docs/clients.rst with dettails about the new mapping logic and added an example

Test Plan:
- Updated Unit tests to catch this is issue
- Tested the changes a real AWS Elasticache Redis cluster fronted by stunnel
  and tested both get, mget and set command with differrent keys

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -12,13 +12,45 @@ host_port_remap
 
 This option exists to enable the client to fix a problem where the redis-server internally tracks a different ip:port compared to what your clients would like to connect to.
 
-The simples example to describe this problem is if you start a redis cluster through docker on your local machine. If we assume that you start the docker image grokzen/redis-cluster, when the redis cluster is initialized it will track the docker network IP for each node in the cluster.
+A simple example to describe this problem is if you start a redis cluster through docker on your local machine. If we assume that you start the docker image grokzen/redis-cluster,
+when the redis cluster is initialized it will track the docker network IP for each node in the cluster.
 
-For example this could be 172.18.0.2. The problem is that a client that runs outside on your local machine will recieve from the redis cluster that each node is reachable on the ip 172.18.0.2. But in some cases this IP is not available on your host system and to solve this we need a remapping table where we can tell this client that if you get back from your cluster 172.18.0.2 then your should remap it to localhost instead. When the client does this it can now connect and reach all nodes in your cluster.
+For example this could be 172.18.0.2. The problem is that a client that runs outside on your local machine will receive from the redis cluster that each node is reachable on the ip 172.18.0.2.
+But in some cases this IP is not available on your host system.To solve this we need a remapping table where we can tell this client that if you get back from your cluster 172.18.0.2 then your should remap it to localhost instead.
+When the client does this it can now connect and reach all nodes in your cluster.
 
-It is also possible to remap the port for each node as well.
 
-Example script
+Remapping works off a rules list. Each rule is a dictionary of the form shown below
+
+.. code-block::
+
+    {
+        'from_host': <Host name on redis server side>, # String
+        'from_port': <Port on the redis server side>, # Integer
+        'to_host': <Host name that the client needs to map to>, # String
+        'to_port': <Port that the client needs to map to> # Integer
+    }
+
+
+Remapping properties:
+
+- This host_port_remap feature will not work on the startup_nodes so you still need to put in a valid and reachable set of startup nodes.
+- The remapping logic treats host_port_remap list as a "rules list" and only the first matching remapping entry will be applied
+- A remapping rule may contain just host or just port mapping, but both sides of the maping( i.e. from_host and to_host or from_port and to_port) are required for either
+- If both from_host and from_port are specified, then both will be used to decide if a remapping rule applies
+
+Examples of valid rules:
+
+.. code-block:: python
+
+    {'from_host': "1.2.3.4", 'from_port': 1000, 'to_host': "2.2.2.2", 'to_port': 2000}
+
+    {'from_host': "1.1.1.1", 'to_host': "127.0.0.1"}
+
+    {'from_port': 1000, 'to_port': 2000}
+
+
+Example scripts:
 
 .. code-block:: python
 
@@ -52,5 +84,43 @@ Example script
     ## Test the client that it can still send and recieve data from the nodes after the remap has been done
     print(rc.set('foo', 'bar'))
 
+This feature is also useful in cases such as when one is trying to access AWS ElastiCache cluster secured by Stunnel (https://www.stunnel.org/)
 
-Pleaes note that this host_port_remap feature will not work on the startup_nodes so you still need to put in a valid and reachable set of startup nodes.
+.. code-block:: python
+
+    from rediscluster import RedisCluster
+
+    startup_nodes = [
+        {"host": "127.0.0.1", "port": "17000"},
+        {"host": "127.0.0.1", "port": "17001"},
+        {"host": "127.0.0.1", "port": "17002"},
+        {"host": "127.0.0.1", "port": "17003"},
+        {"host": "127.0.0.1", "port": "17004"},
+        {"host": "127.0.0.1", "port": "17005"}
+    ]
+
+    host_port_remap=[
+        {'from_host': '41.1.3.1', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17000},
+        {'from_host': '41.1.3.5', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17001},
+        {'from_host': '41.1.4.2', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17002},
+        {'from_host': '50.0.1.7', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17003},
+        {'from_host': '50.0.7.3', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17004},
+        {'from_host': '32.0.1.1', 'from_port': 6379, 'to_host': '127.0.0.1', 'to_port': 17005}
+    ]
+
+
+    # Note: decode_responses must be set to True when used with python3
+    rc = RedisCluster(
+        startup_nodes=startup_nodes,
+        host_port_remap=host_port_remap,
+        decode_responses=True,
+        ssl=True,
+        ssl_cert_reqs=None,
+        # Needed for Elasticache Clusters
+        skip_full_coverage_check=True)
+
+
+    print(rc.connection_pool.nodes.nodes)
+    print(rc.ping())
+    print(rc.set('foo', 'bar'))
+    print(rc.get('foo'))
--- a/rediscluster/nodemanager.py
+++ b/rediscluster/nodemanager.py
@@ -2,6 +2,7 @@
 
 # python std lib
 import random
+import socket
 
 # rediscluster imports
 from .crc import crc16
@@ -9,7 +10,6 @@ from .exceptions import RedisClusterException, RedisClusterConfigError
 
 # 3rd party imports
 from redis import Redis
-from redis._compat import unicode, long, basestring
 from redis.connection import Encoder
 from redis import ConnectionError, TimeoutError, ResponseError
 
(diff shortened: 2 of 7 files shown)


Grokzen:
@appurvj Intersting case you found there. I am very interested in merging this but i need at least two things fixed with this PR. First you need to look at the travis-ci test to see if these changes you introduced caused it to error out, or if you need to rebase your commit on a later commit from this upstream perhaps. Secondly the thing you need to look at is the "usage examples" in the docs that is mentioned in the changelog needs to be updated with a note about this new behaviour and example showing how it works and possibly adding your example scneario you have above where you remap the host and port in different scnearios including your example above.

appurvj:
> @appurvj Intersting case you found there. I am very interested in merging this but i need at least two things fixed with this PR. First you need to look at the travis-ci test to see if these changes you introduced caused it to error out, or if you need to rebase your commit on a later commit from this upstream perhaps. Secondly the thing you need to look at is the "usage examples" in the docs that is mentioned in the changelog needs to be updated with a note about this new behaviour and example showing how it works and possibly adding your example scneario you have above where you remap the host and port in different scnearios including your example above.

@Grokzen  Thanks for your response. Great to see this is worthy of incorporation. I will clean everything up overall and work on addressing the comments. I was not sure how to get a fully reproducible example of this for the `usage examples`(given that it will need multiple environments)  but will just go with dummy values I have used so users have something to reference.

tests/test_cluster_obj.py
```diff
@@ -163,7 +163,7 @@ def test_blocked_commands(r):
     These commands should be blocked and raise RedisClusterException
     """
     blocked_commands = [
-        "CLIENT SETNAME", "SENTINEL GET-MASTER-ADDR-BY-NAME", 'SENTINEL MASTER', 'SENTINEL MASTERS',
+        'SENTINEL GET-MASTER-ADDR-BY-NAME', 'SENTINEL MASTER', 'SENTINEL MASTERS',
```
appurvj, line 166: @Grokzen I have also added this (unrelated) fix for the breaking CI tests since `CLIENT SETNAME` is no longer a blocked command. Please let me know if its better to have this in a separate PR

docs/client.rst
```diff
@@ -52,5 +58,43 @@ Example script
...
+        startup_nodes=startup_nodes,
+        host_port_remap=host_port_remap,
+        decode_responses=True,
+        ssl=True,
+        ssl_cert_reqs=None,
+        # Needed for Elasticache Clusters
+        skip_full_coverage_check=True)
+
+
+    print(rc.connection_pool.nodes.nodes)
```
appurvj, line 97: Note: I have not included comments here cause they are they same as the previous example

appurvj:
@Grokzen I have fixed failing tests and updated docs/client.rst 
Please let me know if anything else needs to be updated.

Grokzen:
@appurvj It looks good with the fixes, i will do a deeper review and just test the feature out in a couple of days and then merge it. I don't see any blocking stuff right now from a quick look at it.

rediscluster/nodemanager.py
```diff
@@ -279,14 +279,18 @@ def remap_internal_node_object(self, node_obj):
...
-
+                    # Try to remap port only when host is remapped. Port cannot be looked at in isolation
+                    # from host level evaluation
+                    # The port value is always an integer
+                    if 'from_port' in remap_rule and 'to_port' in remap_rule:
+                        if remap_rule['from_port'] == node_obj[1]:
+                            # print('remapping port', node_obj[1], remap_rule['to_port'])
+                            node_obj[1] = remap_rule['to_port']
+                # Since these are like evaluation rules, we can stop when the first one "fires"
+                break
```
appurvj, line 290: @Grokzen I was thinking more about this and there might be one (IMO extreme) edge case where this break might cause issues. 
That is if the mapping needs happen on the same pair of hosts but different ports.

So for instance the following configuration:
```
host_port_remap=[
        {'from_host': '2.2.2.2', 'from_port': 18000, 'to_host': '127.0.0.1', 'to_port': 17000},
        {'from_host': '2.2.2.2,  'from_port': 18001, 'to_host': '127.0.0.1', 'to_port': 17001},
        {'from_host': '2.2.2.2', 'from_port': 18002, 'to_host': '127.0.0.1', 'to_port': 17002},
        {'from_host': '2.2.2.2', 'from_port': 18003, 'to_host': '127.0.0.1', 'to_port': 17003},
        {'from_host': '2.2.2.2', 'from_port': 18004, 'to_host': '127.0.0.1', 'to_port': 17004},
        {'from_host': '2.2.2.2', 'from_port': 18005, 'to_host': '127.0.0.1', 'to_port': 17005}
    ]
``` 

- One solution is to move the `break` under the innermost `if` block. So break only if **both** the host and port get re-mapped ,at which point, there is no ambiguity (I like this idea). 
- Or we can remove the `break` all together and and go through all the rules. (In which case, the last matching rule will apply)

Let me know what you think...

appurvj: Or this can be subsequently redesigned to require both host and port under the (from/to) parts of the rules (thus, giving us a unique pointer to each redis process) and go with something like
```
if remap_rule['from_host'] in node_obj[0] and remap_rule['from_port'] == node_obj[1]:
    <only then do re-mapping>
```

Grokzen: I think that if i need to choose between stopping at the first rule or go through all of them and pick the last rule, i would opt not to break at the first hit but to run through all the rules and let the last one win and let the user order all the remapping rules where the last rule is applied. One interesting case you could technically do with that is that you rebind A -> B, but then in the next rule right after you rebind B -> C. I dont know who could really want to do that, but that gives more flexibility to do wonky stuff if one wants :)

appurvj: For now I agree we should remove the break. That should probably keep the edge cases to the minimum. I will push that change here today or tomorrow. We can maybe tackle more complex changes in the next iteration.

appurvj:
@Grokzen  Sorry for the back and forth on this but as I was trying to remove the break, I realized there would still be issues with the approach in some cases. 

So I went a different mapping approach related to your suggestion(at-least my understanding of it).

The new mapping logic uses all available info to check if a remap rule applies. So if host and port info is present , both will need to match for the rule to apply.
This gives us a lot of flexibility. In a rule, users can specify just host mapping or just port mapping or both and all configurations will work.

I have also modified the host matching to explicitly check for equality to avoid bugs like: 
``` python
In [1]: "1.1.1.1" in "11.1.1.1"
Out[1]: True
````

With the current setup all the following rules will be valid and work:
``` python
    {'from_host': "1.2.3.4", 'from_port': 1000, 'to_host': "2.2.2.2", 'to_port': 2000}
    {'from_host': "1.1.1.1", 'to_host': "127.0.0.1"}
    {'from_port': 1000, 'to_port': 2000}
```

I still had to retain the `break` because we should not be processing rules after a remapping has been performed(since that might lead to weird behavior)

I have also updated the docs to reflect. Im feeling pretty good about this change now.
Let me know what you think and (assuming you are ok) what you think would be a good timeline to get this change in..

Grokzen:
Oh right... that is a very strange cornercase indeed that i never thought about. My solution to the ip check out be to run something like `.split(".")` and then just check each item in the list to be the same and if not then it is not a match. In reality tho, i would find it extremley unlikley that someone would bind any IP:s in that fashion from A -> B where that case would be triggered as you have to remember that it basically only allows for leading or trailing additions to the IP. It might screw more on the fourth octet then the first one tho. But that case should be easily fixed anyway.

appurvj:
You are right, then ip check bug would be an ultra-edge case. But is the existing approach of string equality check good enough for now or would you prefer I implement the `.split(".")` based approach?

Grokzen:
No implement the .split(".") approach now, i would change that out later anyway

appurvj:
Sounds good. I will make the change today...

appurvj:
@Grokzen Made the ip addr ccomparison change and added some more validation logic. Please let me know if you think anything is pending

rediscluster/nodemanager.py
```diff
@@ -65,12 +65,30 @@ def _validate_host_port_remap(self, host_port_remap):
...
+            if ('from_port' in item and 'to_port' not in item) or ('from_port' not in item and 'to_port' in item):
+                raise RedisClusterConfigError("Both from_port and to_port must be present in host_port_remap rule if either is defined")
 
-            if ('from_port' in item and 'to_port' not in item) or  ('from_port' not in item and 'to_port' in item):
-                raise RedisClusterConfigError("Both from_port and to_port must be present in remap item")
+            try:
+                socket.inet_aton(item.get('from_host', '0.0.0.0').strip())
+                socket.inet_aton(item.get('to_host', '0.0.0.0').strip())
+            except socket.error:
+                raise RedisClusterConfigError("Both from_host and to_host in host_port_remap rule must be a valid ip address")
```
mrnom, line 82: Host doesn't have to be an IP address like in case of ElastiCache endpoints.

appurvj: Thats a good point, but the redis servers themselves return ips even in Elasticache(going by a test I did). So enforcing using ip addresses will be helpful in catching misconfigurations on that front. In the test that I did with Elasticache, I converted everything to ips and added that in the configuration
 

Having functionality to detect host names and work with host names may be a good feature but IMO may also have edge cases and issues during implementation and (again IMO) may be best done in a separate change and Pull Request

Grokzen: Technically, i could acctually be in facor of implementing two different keywords in the source mapping to chose what kind of remapping you want to have, think like `from_host_ip` and `from_hostname` that would have two different paths through the code. It would not require some magic logic to determine what mode to use and the user can explicitly decide what mode to use and we can implement two different remapping algorithms in each case really. But i would say that this is something that is not super needed to be implemented into this PR right now but can be an addition later on.

appurvj: Yes Agreed.

appurvj:
@Grokzen any estimate of when this might get merged?

appurvj:
@Grokzen just wanted to check in once again and see when this might potentially go through. Please do let me know if anything else needs to be addressed or you me to somehow break down these changes further

Grokzen:
@appurvj Resolved the current conflicts that github found based on my latest changes done in master branch.

rediscluster/nodemanager.py
Grokzen: @mrnom I was thinking about that part about remapping hostnames instead of IP:s makes less sense then to have remapping IP:s. I was thinking that in the AWS ElasticCache case ain't it the purpose of the endpoint manager to handle the IP:s and hostnames to sort them for you. I would love for anyone to actually find a valid use-case where hostname remapping acctually makes any sense in reality over just to add it just to have it but there is no need for it.

Grokzen:
@appurvj I manually merged the branch into the master branch and pushed them up. Github did not allow me to resolve the errors and rebase merge the branch. Consider the issue resolve and committed.

PR #388, opened 2020-09-14
Fix "propergate" instead of "propagate" typo in docs

dan-blanchard:
Was looking through the changes since the 2.0.99 tag and noticed this typo.


Grokzen:
Awesome find @dan-blanchard 

Thank you for this contribution :)

PR #398, opened 2020-09-21
Elasticache example

timvlaer:
I took me a while to get RedisCluster connection working with an AWS elasticache cluster configuration. This example did the trick for me and might help others.


Grokzen:
@timvlaer Looks good, always good with additional new examples. There is also a documentation blog from AWS that i will add in a link to as well later on that describes it in very good details.

PR #402, opened 2020-10-05
allow read from replicas in pipeline

duke-cliff:
Can you merge this and make a 2.1.1 version first?
Using mget is not acceptable as it's adding too much cost to us(3x as the expected cluster size).

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -442,7 +442,7 @@ class RedisCluster(Redis):
         """
         return ClusterPubSub(self.connection_pool, **kwargs)
 
-    def pipeline(self, transaction=None, shard_hint=None):
+    def pipeline(self, transaction=None, shard_hint=None, read_from_replicas=False):
         """
         Cluster impl:
             Pipelines do not work in cluster mode the same way they do in normal mode.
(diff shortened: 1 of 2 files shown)


Grokzen:
@duke-cliff Yes i can make a 2.1.1 releaes after i merge your two PR:s

Grokzen:
Also even if all the current tests actually works, do you know know that the current pipeline algorithm actually works out in basically all pipeline cases even with read-only mode? I have no tests for this and this PR do not include any extension to the tests or at least any manual steps or a validation script to validate that it actually works out as we expect and that it handles a lot of the error scenarios and failover scenarios that actually can happen that the algorithm have to take care of.

duke-cliff:
@Grokzen Yup, it works. The redirect logic is working the same as before, the only difference is when it could get a connection from the slaves for a slot. The rest logic is the same to how you deal with exceptions/redirects(moved/ask) on masters:

image
You can see our CPU was dropped from 16% to 5-6% because of this change.

duke-cliff:
I will have more optimization for this PR coming. testing locally first.

FranGM:
Hi, I noticed that https://github.com/Grokzen/redis-py-cluster/pull/402 is already open for this, is there any way we could help make this happen? We'd be very interested in the ability to be able to send read commands to replicas within pipelines.

jerroydmoore:
When I merged `master` into this branch, the diff went away. I think this PR might be safe to close.

PR #403, opened 2020-10-05
only send readonly to replicas once on new connections

duke-cliff:
image

replicas have more CPU load because of extra commands(2x) than masters.

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -578,7 +578,6 @@ class RedisCluster(Redis):
 
         redirect_addr = None
         asking = False
-        is_read_replica = False
 
         try_random_node = False
         slot = self._determine_slot(*args)
@@ -605,7 +604,6 @@ class RedisCluster(Redis):
                             slot,
                             self.read_from_replicas and (command in self.READ_COMMANDS)
                         )
-                        is_read_replica = node['server_type'] == 'slave'
 
                     connection = self.connection_pool.get_connection_by_node(node)
 
@@ -615,12 +613,6 @@ class RedisCluster(Redis):
                     connection.send_command('ASKING')
                     self.parse_response(connection, "ASKING", **kwargs)
                     asking = False
-                if is_read_replica:
-                    # Ask read replica to accept reads (see https://redis.io/commands/readonly)
-                    # TODO: do we need to handle errors from this response?
-                    connection.send_command('READONLY')
-                    self.parse_response(connection, 'READONLY', **kwargs)
-                    is_read_replica = False
 
                 connection.send_command(*args)
                 return self.parse_response(connection, command, **kwargs)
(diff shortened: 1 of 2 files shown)


Grokzen:
@duke-cliff Most tests fail with this current solution.Seems like the core solution/idea do not work as you expect in your commit there.

Grokzen:
@duke-cliff Also how is this implementation different from the following code lines that exists in the `on_connect` method? https://github.com/Grokzen/redis-py-cluster/blob/6f1fb57c9522ce65ccea7d1a75b1118a32634d02/rediscluster/connection.py#L62

duke-cliff:
@Grokzen I don't think that code works.
1) It's only working for `readonly=True`. However, we need to check `read_from_replicas=True`
2) It's getting the value from `kwargs`: https://github.com/Grokzen/redis-py-cluster/blob/6f1fb57c9522ce65ccea7d1a75b1118a32634d02/rediscluster/connection.py#L51
    But it won't even pass the named params(including read_only/read_from_replcias) from here: https://github.com/Grokzen/redis-py-cluster/blob/2.1.0/rediscluster/client.py#L375 
3) In the ClusterConnection class, I think it might be too late because the connection object itself does not know the `node` information which contains whether the spawned connection is from a master node or a slave node. Unless you have some different thoughts, e.g to pass the node info to every connection object.

Grokzen:
@duke-cliff Actually it will be passed down if i traced the code correct but the path is a bit strange :)

Setting `readonly_mode=True` into the `RedisCluster` class will change the connectionpool class to `ClusterReadOnlyConnectionPool` and it hardcodes `readonly=True` when it calls the super() function to `ClusterConnectionPool` which in turn is bound into `connection_kwargs` in the `ClusterConnectionPool` and those kwargs should be set/sent into each `ClusterConnection` class and it should be tracked for each connection created.

But... the downside with this is that if you set readonly mode, then you will send readonly to all nodes in the cluster when ever any connection is created. Even master nodes which technically is redundant but should not hurt in reality.

I do like this implementation as well, as it properly only sends it to slave nodes and it might work with all subclassed cluster connection pools as well as it moves the logic away from the connection side of things into the generic connection pool which is good, and the hack/workaround in the main client execute logic is removed which i like as well :)

dustMason:
I'm reviewing this thread and the associated branch now – based on what you said above @Grokzen, does this mean we can simply remove [these lines](https://github.com/Grokzen/redis-py-cluster/blob/f0627c91ce23e8784dbc996078428c9bdbacb20b/rediscluster/client.py#L623-L628) and instead rely on the `READONLY` command that gets sent during connection init [here](https://github.com/Grokzen/redis-py-cluster/blob/f0627c91ce23e8784dbc996078428c9bdbacb20b/rediscluster/connection.py#L66), as you describe?

PR #404, opened 2020-10-06
fix(client): Possible UnboundLocalError within execute finally block

mattrobenolt:
I'm not sure what all cases trigger this, but within the finally block,
there is no guarantee previously that the variable `connection` was even
defined.

In out case, we hit a SlotNotCoveredError, which cascaded into an
UnboundLocalError from within the finally block.

This change explicitly declares the `connection` variable so it can
safely be checked later if it needs to be freed.


mattrobenolt:
The test failure seems like a flaky test.

Grokzen:
Yeah i know that happens from time to time in travis. I dont really know exactly why tbh but there is almost in every PR at least one job slize that breaks and i have to manually retrigger the job :)

PR #406, opened 2020-10-09
Reddit patched: Pipeline optimizations

duke-cliff:
I have added the following optimizations, so far we tested the cluster-mode and the result is very stable and much faster than before. Also, the redis load is dropped significantly as well. 

Let me know if you are comfortable with the gevent change, it's good to have but might not be too important to other users also it might not work properly on some platforms. 

1. Allow pipeline to read from replicas
2. Group commands in the pipeline randomly, but the same shard command should just execute in one random node
3. Will retry moved commands using pipeline(stacked_commands)
4. For moved response, it will move replicas with the mater to the new slot
5. Only execute READONLY once per connection
6. Use gevent to execute pipeline commands to different nodes in parallel
7.  Change READ_COMMANDS to set, making the lookup a little bit faster


rediscluster/nodemanager.py
```diff
@@ -257,6 +258,8 @@ def initialize(self):
                 node, node_name = self.make_node_obj(master_node[0], master_node[1], 'master')
                 nodes_cache[node_name] = node
 
+                self.slave_nodes_by_master[node_name] = set()
```
Grokzen, line 261: set? i try to avoid them if possible

rediscluster/nodemanager.py
```diff
@@ -267,6 +270,7 @@ def initialize(self):
                             target_slave_node, slave_node_name = self.make_node_obj(slave_node[0], slave_node[1], 'slave')
                             nodes_cache[slave_node_name] = target_slave_node
                             tmp_slots[i].append(target_slave_node)
+                            self.slave_nodes_by_master[node_name].add(slave_node_name)
```
Grokzen, line 273: Am i guessing right that we need this as a convenience tracking data structure to determine what slaves is for each master as we do not directly has that right now?

rediscluster/pipeline.py
```diff
@@ -15,6 +16,10 @@
 from redis.exceptions import ConnectionError, RedisError, TimeoutError
 from redis._compat import imap, unicode
 
+from gevent import monkey; monkey.patch_all()
```
Grokzen, line 19: I am not a super fan of this addition tho. I think we had a gevent implementation and a threaded implementation way back but we replaced it with the current version that we have in the main execution code path as it was a major overhead to use any kind of threading or parallelization librar/framework to achive some kind of speed improvements which we never saw that we actually gained unless you ran pipelines against a ton of master nodes, not number of commands but just a big number of nodes.

Basically you will have to defend this new implementation why we should add this one over the existing one.

rediscluster/pipeline.py
```diff
@@ -174,71 +179,127 @@ def send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=T
...
+        Send a bunch of cluster commands to the redis cluster.
+
+        `allow_redirections` If the pipeline should follow `ASK` & `MOVED` responses
+        automatically. If set to false it will raise RedisClusterException.
+        """
+        # the first time sending the commands we send all of the commands that were queued up.
+        # if we have to run through it again, we only retry the commands that failed.
+        cmds = sorted(stack, key=lambda x: x.position)
+
+        max_redirects = 5
```
Grokzen, line 242: Number of redirections should be configurable in as a class variable and not hard coded here. Also the main execution loop has 16 TTL rounds before failing out. This has 5 and that is possibly a discrepancy, yes if you get that many moved/ask errors then something is probably more broken then how this code handles the case :)

rediscluster/pipeline.py
```diff
@@ -255,13 +316,19 @@ def _send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=
...
+            log.info("pipeline in slow mode to execute failed commands: {}".format([c.result for c in attempt]))
+
             self.connection_pool.nodes.increment_reinitialize_counter(len(attempt))
+
+            # even in the slow mode, we could use gevent to make things faster
+            events = []
             for c in attempt:
-                try:
-                    # send each command individually like we do in the main client.
-                    c.result = super(ClusterPipeline, self).execute_command(*c.args, **c.options)
```
Grokzen, line 262: Did we not handle the moved and ask errors inside this main execution call? Why move that logic up to inside the pipeline when this call used to handle it all before? Did you find out some new case or reason why this old call was either not working or the new solution covers some cases not previously thought of?

rediscluster/pipeline.py
```diff
@@ -174,71 +179,127 @@ def send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=T
...
+            # build a list of node objects based on node names we need to
+            nodes, connection_by_node = self._get_commands_by_node(cmds)
+
+            # send the commands in sequence.
+            # we  write to all the open sockets for each node first, before reading anything
+            # this allows us to flush all the requests out across the network essentially in parallel
+            # so that we can read them all in parallel as they come back.
+            # we dont' multiplex on the sockets as they come available, but that shouldn't make too much difference.
+
+            # duke-cliff: I think it would still be faster if we use gevent to make the command in parallel
```
Grokzen, line 256: This should be a comment here in the PR, not inside the code with speculation that it might be faster.

rediscluster/pipeline.py
```diff
@@ -174,71 +179,127 @@ def send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=T
...
-        """
-        # the first time sending the commands we send all of the commands that were queued up.
-        # if we have to run through it again, we only retry the commands that failed.
-        attempt = sorted(stack, key=lambda x: x.position)
+        n.read()
 
-        # build a list of node objects based on node names we need to
+    def _get_commands_by_node(self, cmds):
         nodes = {}
+        proxy_node_by_master = {}
```
Grokzen, line 189: I dont really get what this helps with from the docs lines you written about this variable?

Grokzen (changes requested):
Some comments from a initial pass. There is some stuff to discuss around gevent at least, some changes is required to alter back, and some we can talk about/around.

On a side note tho, after diffing into the pipeline code a bit again, i remember back to how complex that piece of code is compared to how i would want it to be way more simpler. I am hoping that in 3.0.0 i can rewrite the node and connection handling good enough to make the pipeline code way simpler compared to today. My dream would be to handle only the commands in the pipeline and just route them to correct node and let the normal client and connection classes handle all the errors and other scenarios that can occur. Will see how that goes down the line tho.

rediscluster/client.py
```diff
@@ -172,7 +172,7 @@ class RedisCluster(Redis):
 
     # Not complete, but covers the major ones
     # https://redis.io/commands
-    READ_COMMANDS = [
+    READ_COMMANDS = set([
```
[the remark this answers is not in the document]
duke-cliff: True, a `frozenset ` would be better here.

rediscluster/nodemanager.py
duke-cliff: I was thinking to lookup the slaves by slave names, it turns out I don't need it eventually. So I think list is perfectly fine here. I will change it.

rediscluster/nodemanager.py
duke-cliff: Exactly, it's for the moved error handling. We should copy the master and the corresponding replicas to the slot.

rediscluster/pipeline.py
duke-cliff: It will be more efficient if we need to talk to multiple shards/nodes in the pipeline.  And for now, it's only used for pipeline commands which should have no impact/overhead on the rest of the logic.

rediscluster/pipeline.py
duke-cliff: Yes, it's an initial PR which I just list out all the things. This would be read from the config for sure.

rediscluster/pipeline.py
duke-cliff: Will remove it.

rediscluster/pipeline.py
```diff
@@ -174,71 +179,127 @@ def send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=T
...
+            # this allows us to flush all the requests out across the network essentially in parallel
+            # so that we can read them all in parallel as they come back.
+            # we dont' multiplex on the sockets as they come available, but that shouldn't make too much difference.
+
+            # duke-cliff: I think it would still be faster if we use gevent to make the command in parallel
+            # the io is non-blocking, but serialization/deserialization will still be blocking previously
+            node_commands = nodes.values()
+            events = []
+            for n in node_commands:
+                events.append(gevent.spawn(self._execute_node_commands, n))
```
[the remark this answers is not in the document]
duke-cliff: When using non-blocking, it only means the thread would not be blocked by the IO, but you still need CPU time to serialize/deserialize data to/from the socket.

So in this case, if you call redis.packed_commands(....) for multiple `NodeCommands` these would be queued in the same thread/process. And with gevent, this would become faster if it runs on multiple cores.

I said this part is nice to have, if you don't want it in the public version, we can talk about whether to give an option to the user or remove it.

rediscluster/pipeline.py
duke-cliff: The proxy_node will be the node to execute the NodeCommands.
My previous PR will just find a random node by get_node_by_slot(slot, True). But this would cause a lot of commands in the same shard would pick different replicas, thus they are not able to be packed efficiently.
So the new logic is:
1) determine the shard of the slot, the shard is the master_node_name
2) find a proxy node for the shard: which is `proxy_node_by_master`
3) group all commands in the same shard to the proxy_node

duke-cliff:
@Grokzen I just updated this PR with these changes:
1) make gevent optional(default: False)
2) make max_redirects configurable(default: 5)
3) make the pipeline smarter to decide whether to use replica node based on if the command is READ_ONLY

rediscluster/pipeline.py
duke-cliff: Because the moved commands can be grouped(stacked) and executed faster. Instead the regular `execute_command` will retry it one by one. I see a significant different by doing this. Also, I read the main logic from redis-go, it's doing the same retries: https://github.com/go-redis/redis/blob/master/cluster.go#L1084

Grokzen:
@duke-cliff I will take these changes out for a test spin and come back to you. It looks good now :)

FranGM:
Hi, I'm curious about the status of this PR, I think we could definitely benefit from some of these changes so we'd love to see it (or some subset of it) merged.

PR #412, opened 2020-10-16
Reduce log level of ClusterConnection creation

TylerLubeck:
Hi! Thanks for setting up all the logging, it's been very helpful.

Would you be up for reducing the log level here? We've found that it's spewing a ton of logs for us. For the time being we've just reduced our output via `logging.getLogger("rediscluster.connection").setLevel(logging.WARN)` but we'd like to keep some of the other info-level logs around.


Grokzen:
Nice cath @TylerLubeck. It will come long with 2.1.1 once the other open PR:s is merged so i can make a new release.

PR #423, opened 2020-12-24
Update connection.py 

noqcks:


Grokzen:
@noqcks Thank you for thiscontribution

Issue #419, opened 2020-11-25
Ensure that the connections obtained from the ConnectionPool are ready to send a comand.


sateeshkumarb:
Looking at the `ConnectionPool` implementation provided by `redis-py` I notice that upon getting a `connection` object from the pool `redis-py` has additional checks to ensure that all the data has been completely read from the socket before the connection object was freed, and the socket has been closed.  (https://github.com/andymccurdy/redis-py/blob/master/redis/connection.py#L1147). 

However in `ClusterConnectionPool` I don't see such checks in `get_connection`. Won't the checks that are done in `get_connection`  of `redis-py` need to be done when we get a connection from the `ClusterConnectionPool`. 

It seems to me, not having such checks could lead to a scenario where a new cluster connection reads the residual data from previous connection. Sorry, I couldn't get a test verifying that not having such a check in `ClusterConnectionPool` leads to problems as it is bit difficult to simulate such a scenario.

Thanks for all your work in creating this module, which is very useful.

Grokzen:
Yeah that is very true and a good find. When i made the first implementation i wanted to use the get_connection() parent method from Redis itself but it was not possible to tie them together so a full custom implementation was needed. I do not like that and i really want to reuse more methods from redis-py itself and i hope that i can get that working for 3.0.0 as i have planned and made progress on with a brand new connectionpool solution.

But in the meantime it should be rather easy to add this feature to the `make_connection` method as that is the one that creates the connection in here https://github.com/Grokzen/redis-py-cluster/blob/master/rediscluster/connection.py#L239

sateeshkumarb:
@Grokzen  Thanks for your response.

Looking at the code it seems to me adding the checks at `make_connection` alone might not be enough. As I see it the checks are needed not only when the new connection is created, and also when any existing connection available from the connection pool is reused (i.e. https://github.com/Grokzen/redis-py-cluster/blob/master/rediscluster/connection.py#L228). It needs to be ensured that the connection that was put back to the pool had no residual data to be read and was properly closed. Or do I miss anything here ?

Grokzen:
Oh yeah you are right it needs to be up there instead, yeah so possibly the get_connection should have it yes.

aajay:
@Grokzen, I have created PR #430 to fix this issue. Could you review? I am not able to add you as a reviewer.

Grokzen:
Fixed in 95fee797dac63501f2933c8a0b46d72dbc5250d9

PR #430, opened 2021-01-29
Add a safety check and improve error handling.

aajay:
- To fix https://github.com/Grokzen/redis-py-cluster/issues/419
   add a safety check to ensure there is no stale data before using the
   connection for executing a command.

- Improve error handling to ensure on any error the connection is disconnected
  before getting released back to the pool.
    
- Set connection to None as on ClusterDownError, the connection pool is reset, so to avoid
  the dangling connection getting added back to the pool.

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -611,6 +611,7 @@ class RedisCluster(Redis):
                     connection = self.connection_pool.get_connection_by_node(node)
 
                 log.debug("Determined node to execute : " + str(node))
+                self._check_ready_to_send(connection)
 
                 if asking:
                     connection.send_command('ASKING')
@@ -665,6 +666,7 @@ class RedisCluster(Redis):
 
             except TimeoutError as e:
                 log.exception("TimeoutError")
+                connection.disconnect()
 
                 if ttl < self.RedisClusterRequestTTL / 2:
                     time.sleep(0.05)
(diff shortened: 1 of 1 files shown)


Grokzen:
@aajay Is this the same code that upstream redis-py uses for this ready check feature?

aajay:
> @aajay Is this the same code that upstream redis-py uses for this ready check feature?

Yes @Grokzen. Added _check_ready_to_send() similar to redis-py. Apart from this, also fixed two other issues as described in the PR.

aajay:
@Grokzen Can you please review this PR?

AmitANetskope:
> > @aajay Is this the same code that upstream redis-py uses for this ready check feature?
> 
> Yes @Grokzen. Added _check_ready_to_send() similar to redis-py. Apart from this, also fixed two other issues as described in the PR.

Reference to similar check added in redis-py library - https://github.com/andymccurdy/redis-py/blob/master/redis/connection.py#L1147

ns-smundle:
@Grokzen Would you be able to help us with this review ?

Grokzen (changes requested):
@aajay I need some kind of steps to validate these changes as there is no added tests inside this PR so w/o that i need to have some way to trigger this manually so i can see what scenario and case that this fixes. w/o that i won't merge this as it is. You can/should also add some tests for this. If this is some kind of fix from upstream redis-py then please port down the tests as well if they have any related to this.

Grokzen:
@aajay I manually implemented this into the code base as i needed to put the check logic elsewhere to mimic the place where redis-py implements it so i added it to the same place and not in the client class

https://github.com/Grokzen/redis-py-cluster/commit/95fee797dac63501f2933c8a0b46d72dbc5250d9

Issue #441, opened 2021-03-09
ClusterPipeline is not exported as part of rediscluster


FranGM:
Hi,

I'm looking into adding clustered redis support to some of our applications and I'm currently hitting a bit of a snag.

Our current usage around `RedisClient` involves subclassing it so we can add instrumentation around our redis usage every time we run a command. We also do the same with pipelines, instead of using `Pipeline` directly we use a subclass of it that adds instrumentation.

We're trying to apply the same logic here and it works well enough with `ClusterRedisClient`, but I just realised that the `rediscluster` package doesn't really export `ClusterPipeline`, (https://github.com/Grokzen/redis-py-cluster/blob/73f27edf7ceb4a408b3008ef7d82dac570ab9c6a/rediscluster/__init__.py#L39-L54) so we can't really subclass it.

We can just do `from rediscluster.pipeline import ClusterPipeline` but it seems cleaner if it's all exported from rediscluster directly.

Is this an intentional design decision or could it just be added to support this use case? I'll get a PR ready just in case it's the latter 😄 

Thanks!

PR #442, opened 2021-03-09
Export ClusterPipeline as part of rediscluster

FranGM:
This will allow users of the library to subclass `ClusterPipeline` as well as `RedisClient`

Fixes https://github.com/Grokzen/redis-py-cluster/issues/441

### DIFF
--- a/rediscluster/__init__.py
+++ b/rediscluster/__init__.py
@@ -23,6 +23,7 @@ from rediscluster.exceptions import (
     MovedError,
     MasterDownError,
 )
+from rediscluster.pipeline import ClusterPipeline
 
 
 def int_or_str(value):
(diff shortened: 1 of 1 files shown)


Grokzen:
@FranGM This looks good an easy to merge and wont harm anyone/anything.

But i am just more curious in how exactly you are patching/sublcassing this really becuase when i look inside redis-py they do not expose this class in their __all__ method and they have no option to change out the pipeline class as a input argument to the `def pipeline(...)` method either.

FranGM:
The short version is that we also have our own subclass of `StrictRedis` which overrides the `pipeline` method (https://github.com/reddit/baseplate.py/blob/develop/baseplate/clients/redis.py#L142-L163) but you're right, they also don't add it to `__all__`, which I actually didn't realise this morning when I was looking at this. Might have needed a second coffee before opening the issue 😅

Grokzen:
@FranGM Thank you for this contribution

PR #448, opened 2021-04-17
Remove arg list training commas (breaks Python2)

rafie:


Grokzen:
Yeah i noticed that -_- i almost rather just remove python2.7 support instead :rofl: 

Thank you for the quick PR @rafie

Issue #400, opened 2020-10-01
Cannot create a read from replicas pipeline.


duke-cliff:
https://github.com/Grokzen/redis-py-cluster/blob/master/rediscluster/pipeline.py#L197

This should pick up the: `self.read_from_replicas`

Also, when we create the pipeline, we should be able to set `read_from_replicas` it from: 
https://github.com/Grokzen/redis-py-cluster/blob/master/rediscluster/client.py#L445-L464

Why we need it?
We often need to read multiple keys(randomly from different nodes), the redis CPU load will be significantly different(60%) when using pipeline vs mget(a get loop in cluster). 
But due to the above bug, we could not use pipeline to read from replicas.

Grokzen:
Good catch @duke-cliff Yes i am aware that pipelines do not support read from replicas today and that is something i want to sort out and fix for 3.0.0 later on. There is even two different readonly/read_from_replica connection pools that are buggy themself that needs to be sorted out. Pipelines in general is often a lower priority to sort out then the main code path through the client. Also note that the logic handling would grow in complexity a fair bit to properly route commands in several new situations. 

Just as an example, if we have 1 master and two replicas per master node. If we want to have readonly mode enabled and the slave node is down, should we route to the master node or the other slave node for reads? Also based on what command you are running you would get different logics like SET style commands must be routed to a master where GET like command that can be offloaded to a pipeline can be routed differently, and you need to track this with MOVED and other cluster specific errors where commands needs to be rerouted for some reason.

duke-cliff:
Yes, I understand the complexity. 

To me, the `readonly` does not mean replicas only, it should route read to all master/replicas. You can find a similar logic in go-redis which supports the cluster mode as well:
https://github.com/go-redis/redis/blob/b965d69fc9defa439a46d8178b60fc1d44f8fe29/cluster.go#L584

In terms of how the pipeline should decide which node to use.
If you want to do something simple, I think we can imply if the pipeline contains any `write` cmds, it will go to masters only.
This can even be set by caller as I suggested above to give a parameter in `cluster.pipeline(readonly=True)`

If you want something smarter you can also read the redis-go logic, it completely routes cmds based on if it's read-only cmd or not.
https://github.com/go-redis/redis/blob/eb28cb64264d03218e5aa8f35eb66a4e4c4ba60b/cluster.go#L1124

kmorrison:
[...]

PR #450, opened 2021-04-20
Support reading from replicas on pipeline executions

FranGM:
This is the smallest possible change that should still satisfy https://github.com/Grokzen/redis-py-cluster/issues/400 

I've been testing it with our own client code and it seems to work as expected.

Something to note about this implementation is that it won't necessarily stick to the same replica for all read operations within the same shard. There are two reasons for this:
* changing this behaviour would complicate the implementation and probably require making the connection pools pipeline-aware somehow as well, or provide a custom connection pool for this use case.
* I actually think the behaviour as implemented is preferable as it will distribute the load across replicas, which is the original goal of this change anyway.

### DIFF
--- a/rediscluster/client.py
+++ b/rediscluster/client.py
@@ -53,6 +53,50 @@ from redis.exceptions import (
     TimeoutError,
 )
 
+# Not complete, but covers the major ones
+# https://redis.io/commands
+READ_COMMANDS = frozenset([
+    "BITCOUNT",
+    "BITPOS",
+    "EXISTS",
+    "GEODIST",
+    "GEOHASH",
+    "GEOPOS",
+    "GEORADIUS",
+    "GEORADIUSBYMEMBER",
+    "GET",
+    "GETBIT",
+    "GETRANGE",
+    "HEXISTS",
+    "HGET",
+    "HGETALL",
+    "HKEYS",
+    "HLEN",
+    "HMGET",
+    "HSTRLEN",
+    "HVALS",
+    "KEYS",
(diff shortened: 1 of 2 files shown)


rediscluster/client.py
```diff
@@ -53,6 +53,50 @@
     TimeoutError,
 )
 
+# Not complete, but covers the major ones
+# https://redis.io/commands
+READ_COMMANDS = frozenset([
```
FranGM, line 58: since I had to move `READ_COMMANDS` to make it available to the pipeline object I ended up converting it into a frozenset (stolen from https://github.com/Grokzen/redis-py-cluster/pull/406) since it does speed up lookups by an order of magnitude 😮

Grokzen:
@FranGM This looks awesome and a much smaller change to get the feature out quicker. Please just dig into it and really confirm it without a doubt that commands is sent out to the replica nodes and i will take this for a run myself and i hope i can get a minor release out within this work week as this should be much faster to verify and get out :)

FranGM:
I can do better than that actually :) I have it running in production right now. Here's a graph showing CPU usage on primary and its three replicas before and after the change 

Screenshot_20210420-224506

Grokzen:
Good enough for me :) I aim for this weekend at the latest to go through it and get out 2.1.3 with this included

Grokzen:
@FranGM Tested it out locally and i also wrote a example script for the readonly pipeline part that i will commit up after this merge to make it more easily testable for anyone who want to try it out.

PR #464, opened 2021-06-30
Make the _created_connections counter works.

iceihehe:
The _created_connections counter is initialized but never increased. Plus 1 when make_connection succeeds.

### DIFF
--- a/rediscluster/connection.py
+++ b/rediscluster/connection.py
@@ -270,6 +270,7 @@ class ClusterConnectionPool(ConnectionPool):
 
         self._created_connections_per_node.setdefault(node['name'], 0)
         self._created_connections_per_node[node['name']] += 1
+        self._created_connections += 1
         connection = self.connection_class(host=node["host"], port=node["port"], **self.connection_kwargs)
 
         # Must store node in the connection to make it easier to track


Grokzen:
@iceihehe This change is not really needed as it is not used in this lib at all more then being reset back to 0 in one specific case/instance when calling the method `reset()`, but this variable is just inherited from the redis-py version of `ConnectionPool` but we do not use it as this ClusterConnectionPool variant use this variable instead `self._created_connections_per_node` where we track the same variable and value but for each individual node instead.

I will not merge this due to this reason above. If you however can show me a practical case for using this variable over the per_node one or find a bug that you can show, then we can reopen this and merge a change, but i do not see it currently.

iceihehe:
It's ok. It's indeed not necessary actually. I need this variable because I just swiched from redis-py to this library and I used this variable as a metrics to reflect the saturation of redis connection pool. Of cause I can use the per node version of this variable instead.

Grokzen:
After some thought, one possible improvement to this feature tho in order to keep the original redis-py code intact is to make this variable into a property on this class that would dynamically calculate this by adding up the connection count from the existing variable and just return that in the same way as redis-py does. If you make a version like that, i will consider it mergeable instead.

iceihehe:
> After some thought, one possible improvement to this feature tho in order to keep the original redis-py code intact is to make this variable into a property on this class that would dynamically calculate this by adding up the connection count from the existing variable and just return that in the same way as redis-py does. If you make a version like that, i will consider it mergeable instead.

I've made a change. Maybe you can have a look at it.

Issue #458, opened 2021-06-02
Possible connection leak in ClusterBlockingConnectionPool


FranGM:
Hey there!

I've been tracking down an issue that I believe could have been caused (or at least made worse) by a possible connection leak in `ClusterBlockingConnectionPool`. Let me try to explain my theory to see if it makes sense.

In [get_connection_by_node](https://github.com/Grokzen/redis-py-cluster/blob/master/rediscluster/connection.py#L495-L539) we pull connections from the queue one by one until we find a connection already established to this node or a `None` (free slot for a connection we can make). 

My concern is with the logic that follows if we run out of items to pull from the queue:

```
            # queue is full of connections to other nodes
            if len(connections_to_other_nodes) == self.max_connections:
                # is the earliest released / longest un-used connection
                connection_to_clear = connections_to_other_nodes.pop()
                self._connections.remove(connection_to_clear)
                connection_to_clear.disconnect()
                connection = None  # get a new connection
            else:
                # Note that this is not caught by the redis cluster client and will be
                # raised unless handled by application code.
                # ``ConnectionError`` is raised when timeout is hit on the queue.
                raise ConnectionError("No connection available")
```

If we have pulled exactly `max_connections` from the pool and all of them are for different nodes then we repurpose one of those connections, otherwise we raise an Exception. Note that when raising this exception we don't put any connections we might have pulled back into the pool.

I think the logic on 
```           if len(connections_to_other_nodes) == self.max_connections:```
is a bug because it implies that all connections are available inside the queue at all times, but connections that are currently being used won't be in the pool, which means there are cases when you have less than `max_connections` elements in the queue (most of the time really), so we will potentially raise an exception when we have connections we could repurpose.

I'm currently testing a fix that would change that line with ``` if len(connections_to_other_nodes) >0:``` but wanted to hear your opinion on this hypothesis before submitting a PR.

Grokzen:
@FranGM I could not say for sure i was not the one who made the blockingconnection pool implementation so to even say anything about it i would have to dig into that code and understand it to be able to answer your question really. I have never really been a fan of this shared style of connection pool as it cause these kinds of issues and other, In 3.0 this kind of error should maybe be gone, not sure with blocking solultion thos, but each node will get their individual connectionpool solution again so each node should sort itself out in the exact same way that redis-py does already so there will be even less reimplementation of code needed which i like much more.

FranGM:
[...]

PR #473, opened 2021-07-19
Stop leaking connections when pool is full

FranGM:
Fixes https://github.com/Grokzen/redis-py-cluster/issues/458

When our connection pool is full (there are no available connections) we
can run into a case where we leak perfectly useable connections. This
essentially reduces the size of our connection pool every time it
happens, as it can slowly leak away our unused connections until we
don't have enough connections available to cope with our workload.

This is particularly problematic when preparing connections for a
pipeline, since we can potentially request (and leak) a large number of
connections in one go.

This change makes sure that before raising an exception if there are no
available connections we return back all connections we've checked to
the pool. In cases where we can't guarantee the connections we'd return
to the pool are in a safe state we just add back `None` objects to the
pool so at least the effective size of the connection pool remains
unchanged.

### DIFF
--- a/rediscluster/connection.py
+++ b/rediscluster/connection.py
@@ -277,7 +277,7 @@ class ClusterConnectionPool(ConnectionPool):
 
         return connection
 
-    def release(self, connection):
+    def release(self, connection, add_to_pool=True):
         """
         Releases the connection back to the pool
         """
@@ -296,7 +296,15 @@ class ClusterConnectionPool(ConnectionPool):
             pass
             # TODO: Log.warning("Tried to release connection that did not exist any longer : {0}".format(connection))
 
-        self._available_connections.setdefault(connection.node["name"], []).append(connection)
+        if add_to_pool:
+            self._available_connections.setdefault(connection.node["name"], []).append(connection)
+        else:
+            # If we don't add it back to the pool it shouldn't count towards the
+            # connection pool, or we'll artificially reduce the maximum size of the
+            # pool
+            self._created_connections_per_node.setdefault(node['name'], 0)
+            if self._created_connections_per_node[connection.node["name"]] > 0:
+                self._created_connections_per_node[connection.node["name"]] -= 1
 
     def disconnect(self):
         """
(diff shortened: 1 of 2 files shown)


rediscluster/connection.py
```diff
@@ -296,7 +296,15 @@ def release(self, connection):
...
             # TODO: Log.warning("Tried to release connection that did not exist any longer : {0}".format(connection))
 
-        self._available_connections.setdefault(connection.node["name"], []).append(connection)
+        if add_to_pool:
+            self._available_connections.setdefault(connection.node["name"], []).append(connection)
+        else:
+            # If we don't add it back to the pool it shouldn't count towards the
+            # connection pool, or we'll artificially reduce the maximum size of the
+            # pool
+            self._created_connections_per_node.setdefault(node['name'], 0)
```
jerroydmoore, line 305:
```suggestion
            self._created_connections_per_node.setdefault(connection.node['name'], 0)
```

rediscluster/pipeline.py
```diff
@@ -203,7 +203,15 @@ def _send_cluster_commands(self, stack, raise_on_error=True, allow_redirections=
             # we can build a list of commands for each node.
             node_name = node['name']
             if node_name not in nodes:
-                nodes[node_name] = NodeCommands(self.parse_response, self.connection_pool.get_connection_by_node(node))
+                try:
+                    nodes[node_name] = NodeCommands(self.parse_response, self.connection_pool.get_connection_by_node(node))
+                except:
+                    # Sommething happened, maybe the pool is full, we need to release any connection
```
jerroydmoore, line 209: typo: 
```suggestion
                    # Something happened, maybe the pool is full, we need to release any connection
```

rediscluster/connection.py
```diff
@@ -538,17 +546,21 @@ def get_connection_by_node(self, node):
...
+    def release(self, connection, add_to_pool=True):
         """
         Releases the connection back to the pool
         """
         self._checkpid()
         if connection.pid != self.pid:
             return
 
+        # In some cases we don't want to add back this connection to the pool but
+        # we still want to free its slot
```
jerroydmoore, line 558:
```suggestion
        # we still want to free its slot (a None represents an available connection
        # pool slot)
```

PR #474, opened 2021-08-05
Make _create_connectioins a property that would dynamically calculate from each node

iceihehe:
Related to #464 , I make this variable into a property on this class that would dynamically calculate this by adding up the connection count from the existing variable and just return that in the same way as redis-py does.

### DIFF
--- a/rediscluster/connection.py
+++ b/rediscluster/connection.py
@@ -384,7 +384,7 @@ class ClusterConnectionPool(ConnectionPool):
         return self.get_master_node_by_slot(slot)
     
     @property
-    def _created_connections(self):
+    def count_all_num_connections(self):
         return sum(self._created_connections_per_node.values())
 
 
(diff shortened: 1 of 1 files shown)


rediscluster/connection.py
```diff
@@ -383,6 +382,10 @@ def get_node_by_slot(self, slot, *args, **kwargs):
         """
         """
         return self.get_master_node_by_slot(slot)
+    
+    @property
+    def _created_connections(self):
+        return sum(self._created_connections_per_node.values())
```
jerroydmoore, line 388: you might want to use self.count_all_num_connections() instead.

jerroydmoore: No, I meant the function `count_all_num_connections` already exists rather than creating a new property that returns the same value.

iceihehe: OK I got it. Sorry for the misunderstood. I'll close this PR.

PR #512, opened 2022-12-05
Redis connection fix

teles7:
Summary: Fixes the exception 'Nonetype object has no attribute disconnect'. The issue happens when the health check runs before execution of a command and during that time the master node shuts down. Since the master went down disconnect is already initiated and the connection is gone. We need to do null check and should not initiate a disconnect if the connection is null.

Fixes: https://jira.team.affirm.com/browse/ONSERV-516


Grokzen:
Hi @teles7 

First i must say that this PR will not be merged due to that this package has been EOL ever since `redis-py` merged this code into their main code line at `4.1.0 (Dec 26, 2021)`. After this i said i would only accept critical security fixes if ever found, otherwise any new changes should be ported up to `redis-py` them self instead. Including these fixes, because i do like the jitter solution, but i am not sure they them self have this problem after all code changes and refactoring they have done by now.

A few general things tho when submitting back code from internal/closed source code repos is that your JIRA link do not work from the outside world. Those things are internal to you and not the rest of the community.

Secondly this PR would not be accepted in the current state anyway. At least it might be squashed before merge to avoid unnecessary merge commits.

Thirdly i don't get why you would update all the versioning to your `affirm` suffix? It would never ever be merged like that.

If you'r company is of the need for this feature inside `redis-py-cluster` right now you will have to make a fork of this code and to maintain these changes your self either internally or under a new name on PyPi as you can't use my package name there.

Do consider however to send this jitter fix only upstream to `redis-py` and see if they have a need for it.

teles7:
Hi @Grokzen , I missed on changing the base branch while creating the PR, sorry. This fix was not meant for redis-py-cluster but for the internal fork that we had.

PR #517, opened 2023-08-24
Do not initialize cluster when cluster slots are empty for node

mayank0903:


Grokzen:
Hi @mayank0903

This repo is no longer active after `redis-py` merged in this code based and took over the maintenance of this code. Since then no changes other then critical security flaws will be handled/merged/released.

My recommendation is that you port this change over to `redis-py` if they are also affected by it. Then you start to move over your codebase from this lib to redis-py if possible. In those cases where that is not possible you will have to run with your own fork including this if you require this change for your own needs

